PR #9, opened 2014-09-03
Revision 1

MartijnR:
Various small corrections, and additions. Should contain nothing controversial. 

Can somebody please review and merge if ok?


MartijnR:
@yanokwa Would you mind having a look and merge if okay? (nothing controversial - just missing info added and edits).

PR #62, opened 2016-12-23
remove banner, part of #46

### LINKED ISSUE #46: Make the status of this documentation clearer
From the discussion at https://github.com/opendatakit/collect/issues/266, this repo does _**not**_ contain the definitive spec for ODK Collect's XForms implementation.  The documentation does have a note at the top, reading

> Work in Progress - Pending peer-review and approval

It would be great to add a link in the same place to the definitive spec at https://bitbucket.org/javarosa/javarosa/wiki/xform

MartijnR:
I think we can do this now because the spec is no more complete and correct than anything else, but totally fine if others want to wait a bit. 

closes #46

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,6 @@ layout: default
 <div class="sidenav clearfix"></div>
 
 <div class="wrap">
-    <h1 style="font-size: 20px; background: orange; padding: 5px;">Work in Progress - Pending peer-review and approval</h1>
 	<article class="post-content">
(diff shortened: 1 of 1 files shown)


lognaturel:
I think the banner is very misleading so I totally agree it needs to go. As we rework the intro, I think it would be good to add something to the effect of "this is a living document, go check out the changes being discussed at https://github.com/opendatakit/xforms-spec/issues".

PR #65, opened 2016-12-23
rewrite of metadata section, closes #47

### LINKED ISSUE #47: Rewrite Metadata section
Not sure where I found preload "instance:. Maybe in JavaRosa code?

MartijnR:
An attempt to bring some order in the metadata chaos.

The current reality is way more flexible than what is written here (mostly because of pyxform's output), but I think that's not ideal and the spec should be more strict. I think this (seemingly) more narrow implementation is currently already supported by all though.

Once this is 'official' we can try to adjust pyxform output to match this (nodeNames and forcing inside a `<meta>` group).

I tried to keep in mind a (far into the) future possible deprecation of the custom jr:preload and jr:preloadParams by replacing them with the standard XForm Actions (as CommCare has done).

It's loosely based on [this](https://bitbucket.org/javarosa/javarosa/wiki/OpenRosaMetaDataSchema) as well.


lognaturel:
I need to chew on this one some more. @yanokwa I think you should also take a look.

_sections/30-bindings.md
```diff
@@ -145,24 +145,59 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
-| property		| email            |      tbc        | Populate with the user's email address if the client has access to this, otherwise 'not supported'.
-| property 		| username         |      tbc        | Populate with username stored in the client. Can be blank if not set or 'not supported'. 
-| property      | phone number     |      tbc        | Populate with phone number of device. Can be blank if not available (e.g. for desktop or tablet) or 'not supported'.
-| property      | simserial        |      tbc        | SIM serial number of phone. May be blank (e.g. for desktop or tablet or 'not supported').
-| property      | subscriberid	   |      tbc        | IMSI of phone prefixed (with imsi: prefix, e.g. imsi:SD655E212). May be blank (e.g., tablets).
+This section describes metadata about _the record_ that is created with the form. Metadata about _the form itself_ (id, version, etc) is covered in the [Primary Instance](http://localhost:4000/#primary-instance) section.
 
+The namespace of the meta block is either the default XForms namespace or "https://openrosa.org/xforms".
```
lognaturel: What's the advantage of mentioning the default XForms namespace option?

lognaturel: Is it because existing engines (JavaRosa, etc) may not have explicit knowledge of the openrosa.org/xforms namespace?

_sections/30-bindings.md
```diff
@@ -145,24 +145,59 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
+            <orx:userID/>
+            <orx:instanceID/>
+        </orx:meta>
+    </data>
+</instance>
+{% endhighlight %}
 
+These meta elements have corresponding `<bind>` elements with a calculation. _Note that these values may be recalculated, e.g. when a draft record is loaded. This could lead to undesirable results especially in case the result is a random value._
+
+Another approach, as mentioned in section [Bind Attributes](#bind-attributes), is to use _preload_ attributes to generate these values at a particular pre-defined time.
```
lognaturel: I think it would be more helpful to have this paragraph before the preload parameter table.

lognaturel (changes requested):
I made a couple of surface comments but I just don't know how to evaluate these changes. I think the organization is certainly better and obviously this is more complete than what's there now.

I don't really understand what you mean when you say that current reality is more flexible. You mean things like the meta block not being required? Anything else?

I'm also afraid I don't quite understand what you mean by "replacing them with the standard XForm Actions".

Maybe with those clarifications I'll feel more comfortable evaluating this!

MartijnR:
Thanks for reviewing!

> I don't really understand what you mean when you say that current reality is more flexible. You mean things like the meta block not being required? Anything else?

Mainly because pyxform places these meta elements in no particular location, e.g. in some group or in the root of the primary instance. Not sure about whether instanceID is currently required by ODK Validate either.

> I'm also afraid I don't quite understand what you mean by "replacing them with the standard XForm Actions".

I was referring to the [XForms "Actions" feature](https://www.w3.org/TR/2006/REC-xforms-20060314/slice10.html) to use `<setvalue>` instead of preload items as [Dimagi is doing]( http://dimagi.github.io/xform-spec/#actions). Fwiw, the advantages I see with this are quite minimal: It's potentially more flexible and doesn't deviate from XForms unnecessarily. There are probably more advantages I don't realize. If we wanted to merge with CommCare, it would be a reason to switch in the future. I know they strongly disliked preload items.

I think what needs to be added, in some form, to this PR is:

What does a combination of preload attribute and the preload parameter attribute do exactly? 
   _It populates a value according to a predetermined fixed formula, when a particular event occurs. Different combinations handle different events (e.g. xforms-ready, xforms-revalidate) and use a different calculation._

What happens if a bind contains a preload parameter as well as a calculation? I think the answer is:
  _Both will populate, and one may overwrite the other. There is probably never a use case, where you'd want to use both._

Does the 'default' value magically populate a meta node if there is no preload attribute and no calculation or no bind at all?
  _Not sure anymore. The original OpenRosa Metadata Schema said it would. For our interest, does JavaRosa populate e.g. meta/instanceID and orx:meta/orx:instanceID automatically without a bind?_

_sections/30-bindings.md
MartijnR: To be honest, I prefer to be strict and require this namespace. The reason I added this is because CommCare still has this flexibility (in their spec). Shall we be strict?

_sections/30-bindings.md
MartijnR: Yes, I see what you mean. Let me re-organize.

_sections/30-bindings.md
lognaturel: Yes, I think so. Then tools can decide whether they want to be fully compliant or more lenient.

But I do want to verify my understanding of something. I know JavaRosa isn't something you're particularly interested in but perhaps you have a general explanation for what's going on -- JR doesn't have any knowledge of the openrosa.org/xforms namespace. So is it just that it currently accepts any namespace for the meta block? I suppose this is also what CommCare does? If you're not sure I can also dig through the code.

lognaturel: Hmm, also, I see that there are a lot of sample forms out there with non-namespaced `meta`. An engine that does not accept it could break on a lot of forms so maybe it's more pragmatic to say something like it should be in the openrosa.org/xforms namespace but backwards compatibility can be offered by accepting non-namespaced or something like that?

lognaturel:
Thanks for the extra context, this is starting to make sense! I really like the additional information you're going to provide.

I tried [meta-bind.txt](https://github.com/opendatakit/xforms-spec/files/713064/meta-bind.txt) and [meta-no-bind.txt](https://github.com/opendatakit/xforms-spec/files/713065/meta-no-bind.txt) (extension changed to txt for GitHub) in ODK Collect and meta/instanceID was not populated without the bind.

_sections/30-bindings.md
MartijnR: Yes, maybe we should just leave it with this flexibility as written. We could still let pyxform generate the namespaced nodes.

Regarding JR, I don't know. One way of dealing with namespaces is to ignore them indeed :). I don't think that's very sound though. I would expect Commcare to not accept any non-allowed namespace for the required meta block (though you could have additional meta blocks in different namespaces - but we probably shouldn't even go there). 

I think the strictness of the meta block structure and namespace has the following advantages:

- The client needs to know where to find and store instanceID and deprecatedID
- The server needs to be able to find the above 2 nodes too for its core functionality
- The server would be able to do analysis across forms on meta data if it knows where to find them.

MartijnR:
Thanks. I made the changes.

P.S. don't tell anybody but fyi Enketo will actually magically populate instanceID and deprecatedID because they are so important to its tools (primary key to store draft records in the browser). Enketo  ignores any calculation or preload attribute on instanceID (and deprecatedID). The pyxform output for instanceID includes a calculation which would replace the instanceID every time the record is loaded.

lognaturel (approved):

PR #68, opened 2016-12-23
added jr:requiredMsg, closes #31

### LINKED ISSUE #31: Add jr:requiredMsg
A surprise addition...

MartijnR:
the wording matches the updated wording of jr:constraintMsg in PR #65

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -41,6 +41,7 @@ The following attributes are supported on `<bind>` nodes. Only the nodeset attri
 | `constraint`| Specifies acceptable answers for the specified prompt with an XPath expression. Will only be evaluated when the node is non-empty.
 | `calculate` | Calculates a node value with an XPath expression.
 | `saveIncomplete` | Specifies whether to automatically save the draft record when the user reaches this question, options `true()` and `false()`. Considered false() if omitted.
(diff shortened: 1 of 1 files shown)


_sections/30-bindings.md
```diff
@@ -41,6 +41,7 @@ The following attributes are supported on `<bind>` nodes. Only the nodeset attri
 | `constraint`| Specifies acceptable answers for the specified prompt with an XPath expression. Will only be evaluated when the node is non-empty.
 | `calculate` | Calculates a node value with an XPath expression.
 | `saveIncomplete` | Specifies whether to automatically save the draft record when the user reaches this question, options `true()` and `false()`. Considered false() if omitted.
+| `jr:requiredMsg` | Specifies the custom message to be displayed when the `required` is violated. Attribute in "http://openrosa.org/javarosa" namespace.
```
lognaturel, line 44: For the namespace, is the goal to specify it at the beginning of lists that are in a certain namespace (here, `constraintMsg`, `preload`, etc are in the same namespace)? I'm wondering whether it might be good enough to mention once at the top.

MartijnR: Yes, could be good to do that. Would you mind if we do that reorganization in a separate PR (to avoid merge conflict with #65)?

lognaturel: I'm currently working on a section between structure and instance about namespaces. I'll approve this now so we get `jr:requiredMsg` and then we can discuss again.

PR #72, opened 2017-01-04
Add description of namespaces used

### LINKED ISSUE #55: Clarify namespaces
@lognaturel writes:

> I know the document pre-supposes a good working understanding of XML but I'm wondering whether it might be worth linking to a quick resource about namespaces along with introducing this (good) change? It might also be good to explain things like the value of having separate OpenRosa XForms and OpenRosa javarosa namespaces as defined in this example.

lognaturel:
I think this is important enough to have its own section but open to other suggestions! Also interested on feedback in the amount of detail.

Closes #55

### DIFF
--- a/_sections/15-namespaces.md
+++ b/_sections/15-namespaces.md
@@ -1,0 +1,11 @@
+---
+title: Namespaces
+---
+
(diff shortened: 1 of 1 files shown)


MartijnR (approved):
Perfect. Thanks!

PR #82, opened 2017-01-31
Add description of mediatype attribute.

### LINKED ISSUE #78: Document mediatype attribute for upload element
Is there a reason this is not explicitly called out?

lognaturel:
I believe this is an accurate description of what [JavaRosa does](https://github.com/opendatakit/javarosa/blob/3eb73b5c8c1d56b9bd635b1aca6a8e9ab1051fa9/core/src/org/javarosa/xform/parse/XFormParser.java#L780). It does not match the [W3C XForms usage](https://www.w3.org/TR/xforms/#ui-upload).

Closes #78.

### DIFF
--- a/_sections/40-body.md
+++ b/_sections/40-body.md
@@ -80,7 +80,8 @@ The following attributes are supported on body elements. Note that most attribut
 | `accuracyThreshold` | For `<input>` with type `geopoint`, `geotrace`, or `geoshape` this sets the auto-accept threshold in meters for geopoint captures. [review]()
 | `value`         | For the `<output>` element to reference the node value to be displayed.
 | `rows`          | Specifies the minimum number of rows a string `<input>` field gets.
-| `mediatype`     | For the `<upload>` element. The string value specifies the kind of media picker that will be displayed. Possible values vary by client. Examples include `image/*`, `audio/*` and `video/*`.
+| `mediatype`     | For the `<upload>` element. The string value specifies the kind of media picker that will be displayed. Unlike in XForms 1.0, only one value can be specified. Possible values vary by client and examples include `image/*`, `audio/*` and `video/*`.
+|`accept`         | For the `<upload>` element. As from the [XForms 2.0 wiki](https://www.w3.org/community/xformsusers/wiki/XForms_2.0#The_upload_Element): "comma-separated list of suggested media types and file extensions used to determine the possible sources of data to upload."
 
 
 ### Appearances
--- a/_sections/40-body.md
+++ b/_sections/40-body.md
@@ -80,6 +80,7 @@ The following attributes are supported on body elements. Note that most attribut
 | `accuracyThreshold` | For `<input>` with type `geopoint`, `geotrace`, or `geoshape` this sets the auto-accept threshold in meters for geopoint captures. [review]()
 | `value`         | For the `<output>` element to reference the node value to be displayed.
 | `rows`          | Specifies the minimum number of rows a string `<input>` field gets.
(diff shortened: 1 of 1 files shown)


MartijnR:
Thanks! 

With respect to deviation, are you referring to not making it a space-separated list?

Note that pyxform did something bad, because of me, and mistakenly used the XForms 2.0 `accept` attribute value for the `mediatype` attribute: https://github.com/XLSForm/pyxform/issues/33 (Where were you back then? :) )

However, since we don't support a space-separated list for mediatype, we may be somewhat lucky in that 'audio/*', 'video/*', 'image/*' are each also valid `accept` attribute values. So in effect our `mediatype` attribute is equal to XForms 2.0 `accept` attribute.

Do you think we should mention that and add a [link](https://www.w3.org/community/xformsusers/wiki/XForms_2.0#The_upload_Element)? This PR is fine as-is as well though.

lognaturel:
Yes, the discrepancy I was referring to is specifying a list vs a single value. JavaRosa currently only accepts a single value for `mediatype` and that value is used to display a specific Collect widget. Should the spec say that a comma-separated list can be specified (as in `accept`)? If so, what should "audio/*,video/*" do, for example? That's a useful list as a filter for picking an existing media file but not so much for offering an opportunity to capture new media. If Enketo allows for a list in `mediatype`, what does it do with that list?

MartijnR:
I think it's simpler in the browser because we have no control over this so it requires no deliberation. Enketo copies the mediatype value and places the unchanged value as the HTML `accept` attribute of a file input (like XForms 2.0). The combination "audio/\*,video/\*" has no issues as far as I can see. [See demo](http://codepen.io/MartijnR/pen/VPXvyy). 

On a desktop the browser/OS file picker will grey out all non-image and non-video files. On a mobile device most browsers will give the user the option of using a file browser or apps that can produce either video or image files or both.

I think this built-in behaviour is perfect though. I actually wouldn't know how I would want to improve this (if it were possible).

So I guess the question is, do we want to support complex filters (lists) such as done for type=file (generic file input, but not .exe etc) (and future type=doc, type=spreadsheet each encompassing a list of several formats)?

If so, we could:
1. support space-separated list as in XForms 1.0 (and change pyxform output for type=file, not a big deal)
2. support the `accept` attribute as in XForms 2.0, as well as the `mediatype` attribute but without lists. If `accept` is present, `mediatype` will be ignored (And then change pyxform output to `accept` at least for type=file, but maybe for the others as well)
3. support a hybrid (deviating) `mediatype` attribute that accepts `accept` values as described in XForms 2.0.

Not that it matters, but none of these are significant for Enketo to adopt whilst maintaining backwards compatibility for old forms. I need to ponder to figure out which one I'd prefer.

lognaturel:
I like your option 2 the best. I find it a nice balance of codifying existing behavior and getting closer to something that's W3C spec compliant.

MartijnR:
Ok, cool. I'll add an issue to pyxform to change type=file output to `accept`. -> https://github.com/XLSForm/pyxform/issues/101

Do you want to add `accept` to this PR or (me to) do this separately?

lognaturel:
I can do it! So in general, it seems like [this spec](https://www.w3.org/TR/xforms20/#The_upload_Element) is obsolete and replaced by [this wiki](https://www.w3.org/community/xformsusers/wiki/XForms_2.0#The_upload_Element)? I guess we're not the only ones with versioning challenges... 😁

MartijnR:
Yes! 👍

lognaturel:
Hmm, I totally forgot to mention that `mediatype` is ignored if `accept` is also specified. Do you think that should be in the description for `mediatype` or after the table? Let me know and I'll issue another PR.

MartijnR:
Ah yes, sorry I missed that too. I think it's best to put it under mediatype.

PR #84, opened 2017-02-01
Improve intro

lognaturel:
I realized the first sentence could be read as "The ODK community uses the W3C XForms spec. The ODK XForms is a subset of the W3C XForms spec." Hopefully this is clearer.

This also fixes the ordered list which wasn't being displayed because of a missing newline.


MartijnR:
👍

PR #113, opened 2017-05-04
Added contains(), starts-with() and ends-with functions, closes #105

### LINKED ISSUE #105: Add support for contains(), starts-with(), ends-with() functions
starts-with() and contains() as in XPath 1.0. ends-with() is XPath 2.0 (I believe)

This will provide powerful search capabilities for secondary instances (including external). Especially after #88, it would greatly increase the search capabilities.

E.g. finding items that do not contain a certain string (as requested [here](https://groups.google.com/forum/#!topic/opendatakit/lMams3XcaNc)) with `not(contains(..))`

Currently, we can only really search XML instances for exact value matches.

MartijnR:


lognaturel:
I really don't understand how `starts-with` was added to the 1.0 spec without `ends-with`! 🤔

MartijnR:
puzzling indeed!

PR #116, opened 2017-05-04
Describe the range control and its attributes.

### LINKED ISSUE #109: Add W3C XForms range control
Add support for the `range` control as described [here](https://www.w3.org/TR/xforms/#ui-range). Initially we could omit the `incremental` property since I don't believe any tools in the ODK XForms ecosystem supports events currently.

XLSForm/pyxform#79 takes this up from the XLSForm angle but we probably don't need to think too much about that until JavaRosa implements it.

@spatialdude

lognaturel:
Closes #109


MartijnR:
Great!

PR #144, opened 2017-09-15
Define what a primary instance is in the primary instance section.

### LINKED ISSUE #143: Clarify how the "primary instance" is identified
As far as JavaRosa is concerned, the "primary instance" is the top-most instance. @MartijnR is this also the case in Enketo? Can we codify this? I think it's an important thing to have clearly defined!

lognaturel:
This is mostly to make myself feel better for not finding what I was looking for and filing #143. Feel free to close if deemed redundant. 

I jumped straight to the "primary instance" section when looking for the strict definition so I can imagine it could help someone else to have it there.

Closes #143

### DIFF
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -6,7 +6,7 @@ A `<model>` can have multiple instances as childnodes. The first and required `<
 
 ### Primary Instance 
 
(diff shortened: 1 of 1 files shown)


MartijnR:
Great! Thank you!

PR #149, opened 2017-10-01
Added section on encryption

### LINKED ISSUE #133: Add encrypted form specification
In building out encrypted forms from @clint-tseng noted that how to encrypt forms isn't described in the spec. Seems like it should be. https://opendatakit.org/help/encrypted-forms has the details.

> The other required element to make this form an encrypted form is the submission tag. Within this tag, the method attribute should always be form-data-post. The action attribute should be the url to which the submission should be posted; this is the ODK Aggregate website url with Aggregate.html replaced by submission. Finally, what identifies the form as an encryted form is the presence of a base64RsaPublicKey attribute. This should be the base64 encoding of the RSA public key that ODK Collect uses to encrypt the symmetric encryption key it creates to encrypt a finalized instance of this form (a different symmetric encryption key is created for every finalized form).

In my testing, the only thing that is required is a `submission` tag with the `base64RsaPublicKey` attribute. @mitchellsundt it seems like `method` and `action` are should have and `base64RsaPublicKey` is a must have. Is that correct? What was the rationale behind having a method and action?

Related: [XForms Submission spec](https://www.w3.org/TR/xforms/#submit)

Divya063:
Fixes #133

### DIFF
--- a/_sections/91-encrypted-forms.md
+++ b/_sections/91-encrypted-forms.md
@@ -1,18 +1,34 @@
 ---
 title: Encryption 
 ---
-Encrypted forms provide a mechanism to keep your data private even when using http: for communications (e.g., when you do not have an SSL certificate or https: is not available).
+Encrypted forms provide a mechanism to keep finalized data private even when using **http:** for communications (e.g., when SSL certificate is not there).It provides security for the duration in which the data is stored on a device and on the server.
 
 ### Form Definition
+
+Encrypted form definitions must have an explicit `<submission/>` element with the following attributes:
+
+- `method`: should always be set to `form-data-post`.
+- `action`: should be the url to which the submission should be posted; this is the server url.  
+- `base64RsaPublicKey`:base64 encoding of an RSA public key. The corresponding private key will be needed to decrypt submissions.
+
+The client generates a different symmetric encryption key for each finalized form and uses it to encrypt the submission and all media files. The `base64RsaPublicKey` is used to encrypt the symmetric key with the **RSA/NONE/OAEPWithSHA256AndMGF1Padding** algorithm.
 
 Here is an excerpt used in an encrypted form:
 
 {% highlight xml %}
+<instance>
+      <sample id="sample-v1.0">
+        <orx:meta>
+           <orx:instanceID/>
+        </orx:meta>
+        <name/>
+      </sample>
+</instance>
 <submission method="form-data-post"
                 action="https://my-opendatakit.appspot.com/submission"
                 base64RsaPublicKey="MIIBIjANB...JCwIDAQAB" />
 {% endhighlight %}
 
-The form encryption is governed by the settings on the [Settings Worksheet](https://opendatakit.org/help/form-design/xlsform/#settings_ws).Encrypted forms must specify a *submission_url* and a *public_key* on this worksheet.If both are specified, XLSForm will generate a encrypted-form definition.Encrypted form definitions must include a **unique OpenRosa instanceID** and have an explicit `<submission/>` element.Within this tag, the method attribute should always be `form-data-post`. The action attribute should be the url to which the submission should be posted; this is the server url. Finally, what identifies the form as an encryted form is the presence of a *base64RsaPublicKey* attribute. This should be the base64 encoding of the RSA public key that client uses to encrypt the symmetric encryption key it creates to encrypt a finalized instance of this form (a different symmetric encryption key is created for every finalized form).
+
 				
 
--- a/_sections/91-encrypted-forms.md
+++ b/_sections/91-encrypted-forms.md
@@ -1,0 +1,18 @@
+---
+title: Encryption 
+---
+Encrypted forms provide a mechanism to keep your data private even when using http: for communications (e.g., when you do not have an SSL certificate or https: is not available).
+
+### Form Definition
+
+Here is an excerpt used in an encrypted form:
+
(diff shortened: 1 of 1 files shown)


_sections/91-encrypted-forms.md
```diff
@@ -0,0 +1,18 @@
...
+
+{% highlight xml %}
+<submission method="form-data-post"
+                action="https://my-opendatakit.appspot.com/submission"
+                base64RsaPublicKey="MIIBIjANB...JCwIDAQAB" />
+{% endhighlight %}
+
+The form encryption is governed by the settings on the [Settings Worksheet](https://opendatakit.org/help/form-design/xlsform/#settings_ws).Encrypted forms must specify a *submission_url* and a *public_key* on this worksheet.If both are specified, XLSForm will generate a encrypted-form definition.Encrypted form definitions must include a **unique OpenRosa instanceID** and have an explicit `<submission/>` element.Within this tag, the method attribute should always be `form-data-post`. The action attribute should be the url to which the submission should be posted; this is the server url. Finally, what identifies the form as an encryted form is the presence of a *base64RsaPublicKey* attribute. This should be the base64 encoding of the RSA public key that client uses to encrypt the symmetric encryption key it creates to encrypt a finalized instance of this form (a different symmetric encryption key is created for every finalized form).
```
MartijnR: We keep XLSForm references out of this spec since XLSForm is just one (great) way of generating XForms, and to avoid confusion. So only describe `<submission>`, `method`, `action`, `base64RsaPublicKey` here.

MartijnR:
Thanks! I will add some comments, and hopefully @lognaturel can review as well.

_sections/91-encrypted-forms.md
MartijnR: **unique OpenRosa instanceID* is required for all forms, so probably not needed here.

MartijnR: Also describe that attachments (file uploads) are encrypted.

_sections/91-encrypted-forms.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Encryption 
+---
+Encrypted forms provide a mechanism to keep your data private even when using http: for communications (e.g., when you do not have an SSL certificate or https: is not available).
```
MartijnR: In addition, it provides security for the duration in which the data is stored on a device and on the server.

_sections/91-encrypted-forms.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Form Definition
+
+Here is an excerpt used in an encrypted form:
+
+{% highlight xml %}
+<submission method="form-data-post"
```
MartijnR: add the primary instance as well to show where the submission element goes

_sections/91-encrypted-forms.md
lognaturel: Agreed. I would also remove "you" to match the style of the rest of the spec -- "keep data private", "when there is no SSL"

_sections/91-encrypted-forms.md
lognaturel: Good idea. I suggested only including the `submission` block but it's true that it's not clear where it goes. I think adding 
```
<instance>
...
</instance>
```
before it would be sufficient.

_sections/91-encrypted-forms.md
lognaturel: Yes, nothing XLSForm related should go in here. Additionally, I would move the narrative before the example. I would go with something like:

Encrypted form definitions must have an explicit `<submission />` element with the following attributes:
- `method`: always set to `form-data-post`
- `action`: the url of the server to send this submission to
- `base64RsaPublicKey`: base64 encoding of an RSA public key. The corresponding private key will be needed to decrypt submissions. 

The client generates a different symmetric encryption key for each finalized form and uses it to encrypt the submission and all media files. The `base64RsaPublicKey` is used to encrypt the symmetric key with the RSA/NONE/OAEPWithSHA256AndMGF1Padding algorithm. 

===

I'm not sure about the last paragraph as it's starting to get into non-XForms details. Perhaps it should live in the API docs which would also describe the manifest.

@Martijn, do you think `method` should be part of this spec? Is it likely to have any other valid values? W3C XForms uses it for HTTP method as described in https://www.ibm.com/support/knowledgecenter/SSS28S_8.2.0/XFDL_Specification/xfdl_r_xforms_submission.html But I don't think any tool in the ODK ecosystem supports those.

lognaturel (changes requested):

_sections/91-encrypted-forms.md
MartijnR: I would want to say "finalized data" actually, because that's a rather important thing.

_sections/91-encrypted-forms.md
MartijnR: I think `method` should be in there (but it should actually have value `"post"` in hindsight). Your question leads me to suspect that ODK Collect is not actually reading/using this attribute. In that case, we could change it to `"post"` perhaps....

The last paragraph (about the encryption algorithm) is good to include here, I think.

Divya063:
@lognaturel @MartijnR Thanks for the review,  applied the requested changes.

lognaturel (approved):

lognaturel:
Thank you @Divya063 and apologies for the delay in getting it merged!

PR #154, opened 2017-11-15
Describe the format produced by decimal-date-time

### LINKED ISSUE #45: decimal-date-time() and decimal-time()
Expand spec by explaining _what kind of number_ these convert to. Seconds? Milliseconds?

### LINKED ISSUE #714: Create Online Filepicker
To be used in online-only views.

It may be best to use an alternative FileManager for this that has the same methods as the fancy offline-capable FileManager.

lognaturel:
Closes enketo/enketo#714

Confirmed at https://forum.opendatakit.org/t/calculating-duration/2959/3 I think this is important to specify so implementations can be compatible.


MartijnR:
https://github.com/opendatakit/xforms-spec/issues/45

_sections/30-bindings.md
```diff
@@ -112,8 +112,8 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
 <a id="fn:boolean-from-string" href="#fn:boolean-from-string">`boolean-from-string(string arg)`</a> | Returns true if arg is "true" or "1", otherwise returns false.
 <a id="fn:not" href="#fn:not">`not(boolean arg)`</a> | As in [XPath 1.0](http://www.w3.org/TR/xpath/#function-not).
 <a id="fn:number" href="#fn:number">`number(* arg)`</a> | As in [XPath 1.0](http://www.w3.org/TR/xpath/#function-number).
-<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | Converts dateTime value to a number.
-<a id="fn:decimal-time" href="#fn:decimal-time">`decimal-time(time value)`</a> | Converts time value to a number.
+<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | Converts dateTime value to a number in the format used by Excel. The number before the decimal marker is the number of days since January 1, 1970. The number after the decimal marker represents the time as a fractional date.
```
MartijnR: Wow. I would have never ever guessed this! The number after the decimal marker means 3:00am is x.25 and noon is x.50?

_sections/30-bindings.md
```diff
@@ -112,8 +112,8 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
+<a id="fn:decimal-time" href="#fn:decimal-time">`decimal-time(time value)`</a> | Converts time value to a number representing the number of days since January 1, 1970.
```
MartijnR: I do not understand how this is possible... If the parameter was a dateTime value it would make sense, but then it would be the same as date-time.

_sections/30-bindings.md
MartijnR: We should probably add "UTC" after 1970. I wonder if it would be easier to just leave out the after-decimal part. It seems unnecessary especially if we also take timezone into consideration.

MartijnR (changes requested):

_sections/30-bindings.md
lognaturel: Ohhhhhh. Wow. I'm sorry, for some reason I convinced myself this was `decimal-date`. 😖

_sections/30-bindings.md
lognaturel: When you say leave it out, you mean not explicitly explain it because it's implied?

_sections/30-bindings.md
lognaturel: I'll need to double check to make sure I actually understand the behavior and then fix this. Will probably not be today.

lognaturel:
@MartijnR I've confirmed the desired behavior of `decimal-time` and hopefully have documented it accurately this time. I didn't understand your comment about omitting the after-decimal part in the `decimal-date-time` description.

_sections/30-bindings.md
```diff
@@ -112,8 +112,8 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
 <a id="fn:boolean-from-string" href="#fn:boolean-from-string">`boolean-from-string(string arg)`</a> | Returns true if arg is "true" or "1", otherwise returns false.
 <a id="fn:not" href="#fn:not">`not(boolean arg)`</a> | As in [XPath 1.0](http://www.w3.org/TR/xpath/#function-not).
 <a id="fn:number" href="#fn:number">`number(* arg)`</a> | As in [XPath 1.0](http://www.w3.org/TR/xpath/#function-number).
-<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | Converts dateTime value to a number.
-<a id="fn:decimal-time" href="#fn:decimal-time">`decimal-time(time value)`</a> | Converts time value to a number.
+<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | Converts dateTime value to a number in the format used by Excel. The number before the decimal marker is the number of days since January 1, 1970 UTC. The number after the decimal marker represents the time as a fractional day.
```
MartijnR: Hi @lognaturel, I couldn't find your earlier comment to respond to. What I meant was, the part about after the decimal marker seems confusing especially if we take into account timezone conversion. Would something like this be better perhaps (there is no separate before-decimal-marker and after-decimal-marker calculation I think):

`Converts dateTime value to the number of days since January 1, 1970 UTC`

_sections/30-bindings.md
```diff
@@ -112,8 +112,8 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
+<a id="fn:decimal-time" href="#fn:decimal-time">`decimal-time(time value)`</a> | Converts time value to a number representing a fractional day. For example, noon is 0.5 and 6pm is 0.75.
```
MartijnR: Where this becomes unclear is how to convert time zones. My guess is it should be converted to UTC and use negative values if you move backwards past midnight UTC, and use a value larger than 1 if you move forward past midnight UTC.

MartijnR (changes requested):

_sections/30-bindings.md
lognaturel: I feel strongly that there should be no negatives. In clock time, 00:00 and 24:00 are the same thing. Similarly, in this scheme 0 and 1 are the same.

The behavior that has existed since the introduction of this function in JavaRosa is to ignore timezone. I think in this case it does make sense because the function exists to allow calculations on raw times. I think the primary use case is subtracting one time from another to get an interval. Things are a lot easier to reason about without having to consider timezone conversions.

_sections/30-bindings.md
lognaturel: That works for me. `decimal-time` explains the fractional day idea if it's unclear. I'll still include the fact that it matches the Excel format because I think it helps explain why this was picked over more traditional epoch representations.

For old comments, there's a little "Show outdated" toggle somewhere up there ⏫ . 😊

MartijnR: Ah! Thanks. I see the old comments now!

Sorry, I was not aware of any other decimal epoch representations (apart from using seconds or milliseconds instead of days). So maybe mentioning Excel makes sense if there is scope for confusion.

lognaturel: Seconds or milliseconds is what I know. I hadn't seen the fractional day one before but that doesn't mean it's not common! When you wrote "Wow. I would have never ever guessed this! The number after the decimal marker means 3:00am is x.25 and noon is x.50?" up there ⏫ I thought you might have been surprised by it too!

_sections/30-bindings.md
MartijnR: I would have agreed with you until about a month ago when I learned that [ODK Collect's time format included the timezone](https://github.com/enketo/enketo-core/issues/463) and that Enketo was embarrassingly not following the spec (rectified now). Stripping timezone info away in decimal-time() feels wrong to me. Isn't the addition of timezones in JavaRosa's time format recent also?

@alxndrsn, @tiritea, any opinions?

lognaturel: Ooh, actually, I was wrong about the existing behavior. It's not stripped. Time is converted to the current timezone, as in Excel (https://en.wikipedia.org/wiki/Decimal_time#Fractional_days -- "Microsoft Excel uses the local time zone of the computer").

I'm currently in UTC-8. If I have a timestamp of 16:00:00.000-8 and call `decimal-time` on it, I get 0.6666666666678793. If I have a timestamp of 16:00:00.000-12, I get 0.8333333333321207 (16:00-12 is 20:00-8).

It does make sense to convert to one consistent timezone now that I think about it. Definitely interested in what others think. Timezones give me headaches.

_sections/30-bindings.md
MartijnR: :) Yes, I was initially thrown off by the description, thinking there was a different meaning of the value before and after the decimal marker. But then realized there isn't. So it's just the unit that differs.

lognaturel:
I have clarified that `decimal-time` should always be in the device's timezone. Hopefully that behavior is the expected one and the docs are now complete enough to ensure compatible implementations. @MartijnR @alxndrsn @tiritea

alxndrsn:
I don't think we're actively using this function, so you won't get any complaints from this direction 😁

MartijnR (approved):

MartijnR:
Thank you!

tiritea:
Bit late to the party, but... @lognaturel 

> I have clarified that decimal-time should always be in the device's timezone.

What is the rationale for this choice/decision? vs, say, making all `decimal-times` UTC based. [not disagreeing per se, just asking why...]

lognaturel:
@tiritea I wasn't privy to how the decision was made but I think that it was to match what Excel does.

tiritea:
There appears to be some prior art here, wrt [W3C XML Schema Definition](https://www.w3.org/TR/xmlschema11-2/#vp-dt-timezone); specifically

> ...All of the value spaces for dates and times described here represent moments or periods of time in Universal Coordinated Time (UTC).

basically, my reading of it is that for _their_ dateTime (and related) primitive data types, when the optional timezone information is omitted, it assumes UTC (!). Not sure if we want to follow W3C's lead, or Microsoft's... ;-)

lognaturel:
Tough choice! I agree with your reading. 

I think these two functions, `round` with two arguments and some others were added to match Excel so it seems fine for those to follow those rules for those special ones. Also, since these functions are only useful to do date/time math, I don't think it's too important which timezone things are converted to, only that there is a consistent standard.

alxndrsn:
> Also, since these functions are only useful to do date/time math, I don't think it's too important which timezone things are converted to, only that there is a consistent standard.

Is it obvious which is more consistent out of UTC and local time?

If calculating the difference between two times which cross a boundary for daylight saving changes in local time, I can think of situations in which both might be useful.

In general though, I _suspect_ using UTC will give more predictable behaviour and therefore less surprises, and less bugs.

tiritea:
> Is it obvious which is more consistent out of UTC and local time?

Unfortunately, I'd probably have to side with UTC here. _Especially_ when taking the difference between two times, which in effect **must** be normalized against common timezone

The only advantage I see to local time is to save a step (on the client) when having to convert UTC to local time for immediate _display_ purposes. In pretty much all other cases, having an unambiguous time - ie referenced against an explicit (or implicit) timezone - would be more preferable, and predictable. Local times are problematic, especially when we are considering _mobile_ [sic] devices; its really only preferable for display/UI interaction.

lognaturel:
I don't disagree with you in principle @tiritea @alxndrsn. I think these cases where we're finally documenting things that have been in one implementation or another for years are a little tricky. The extreme cases are pretty clear:
- matches W3C XForms/XPath/some other existing spec -- it directly gets documented as-is
- totally egregious abuses of syntax -- keep those undocumented at the spec level and flag by implementations as things to eventually redress

But what about something like this which is somewhere in the middle? It's an addition from outside the base specs. It's presumably mapped to an unrelated system's standards (Excel). It's maybe not ideal but I would say definitely defensible and not limiting. In that case, I tend to think matching an existing implementation is the least disruptive path to follow.

tiritea:
> I don't disagree with you in principle

Well, since @alxndrsn has already admitted he's not actually using any of this junk - and neither am I - yer probably perfectly free to ignore these 'armchair quarterbacks'. 😏 

But seriously, if nobody is actually using (ie relying on) it at present - and its not too much (re?)work - I do think UTC-based could avoid problems down the line... Just my $0.02.

PR #155, opened 2017-12-01
Add events and actions

lognaturel:
My intent was for this to provide enough information to get started by leaning on existing W3C documentation.

I considered documenting different attributes but in the end I thought providing two clear examples was enough information when combined with the W3C links.

I named the section "events and actions" with spaces so I had to add a rule to replace spaces with dashes when generating section title ids.

I chose to put the section where I did because it seems more advanced than the ones that precede it and depends on some of them.

### DIFF
--- a/_sections/65-events-and-actions.md
+++ b/_sections/65-events-and-actions.md
@@ -1,0 +1,43 @@
+---
+title: Events and Actions
+---
+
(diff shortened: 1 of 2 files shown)


MartijnR (approved):
This looks perfect to me. Thanks for writing this up and sorry for taking so long to review it.

PR #159, opened 2018-02-19
Added documentation about form_logo.png, closes #34

### LINKED ISSUE #34: Include form_logo.png feature in documentation

MartijnR:
Not crazy about this, but I think we should include it.


lognaturel:
This is no longer supported by ODK Collect -- users are instead encouraged to design their own landing/branding screens explicitly using media includes. Perhaps it should just say "this image MAY be shown by the client"

MartijnR:
I think it makes sense to not have this feature at all. The only issue I have with the proper approach that you started using is that I'd like a 'note' with a media label that serves as logo to be display above the form title, but that's pretty minor so I'll get over that.

lognaturel:
I'm happy either way! I do think specifying a logo makes more sense for a web client like Enketo so if you prefer to document it that works for me.

MartijnR:
Let's be strict. It's a good feature, but poorly implemented. True, that it makes more sense for a web client indeed.

PR #164, opened 2018-03-07
Describes format-date-time, closes #20

### LINKED ISSUE #20: add format-date-time()

MartijnR:


_sections/30-bindings.md
```diff
@@ -117,7 +117,8 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
+<a id="fn:format-date-time" href="#fn:format-date-time">`format-date-time(dateTime value, string format)`</a> | Returns the provided dateTime value formatted as defined by the format argument using the following identifiers:<br/>`%Y`: 4-digit year<br/>`%y`: 2-digit year<br/>`%m` 0-padded month<br/>`%n` numeric month<br/>`%b` short text month (Jan, Feb, etc)\*<br/>`%d` 0-padded day of month<br/>`%e` day of month<br/>`%H` 0-padded hour (24-hr time)<br/>`%h` hour (24-hr time)<br/>`%M` 0-padded minute<br/>`%S` 0-padded second<br/>`%3` 0-padded millisecond ticks<br/>`%a` short text day (Sun, Mon, etc).\* <br/>\* If form locale can be determined that locale will be used. If form locale cannot be determined the locale of the client will be used (e.g. the browser or app).
```
lognaturel: How about saying "... using the same identifiers as [link](format-date) plus the following:" Maybe I'm a bit eccentric but my reaction when I see a possibly duplicated list like this is to very carefully compare them to make sure there's no trick. Having the date identifiers in one place would give me confidence they're meant to be the same.

MartijnR: Thanks. Makes sense. Will do!

lognaturel (approved):

PR #167, opened 2018-03-13
Added randomize function, closes #51

### LINKED ISSUE #51: Proposal: Randomizing nodesets
Add support for a custom randomize() function, that takes a single nodeset as parameter.

[earlier discussion on dev forum](https://groups.google.com/forum/#!searchin/opendatakit-developers/randomize%7Csort:relevance/opendatakit-developers/plutkYCqWB4/QAcV1SZsBQAJ)

As in any XPath function, there would be no restrictions on its use, but the use case that triggered this proposal is to randomize select and select1 options like this:

```xml
<select1 ref="/form/state">
    <itemset nodeset="randomize(instance('choices')/states/state)">
        <label ref="name"/>
        <value ref="value"/>
    </itemset>
</select1>
```

MartijnR:
I wouldn't be surprised if this requires some editing... I got pretty lost with the seeded shuffle and how to describe that in a way that would allow implementors to develop the same RNG in multiple languages.


lognaturel (approved):

lognaturel:
I don't know, it makes sense to me!

MartijnR:
Once somebody starts implementing the seeded part in a second language, the questions may prompt further elaboration. I wouldn't really be able to implement the seeded part with this description at the moment (in a way that would guarantee the same result in other implementations).

lognaturel:
The Java implementation will use `Collections.shuffle` which implements Durstenfeld's version of the Fisher-Yates algorithm: https://docs.oracle.com/javase/9/docs/api/java/util/Collections.html#shuffle-java.util.List-java.util.Random-

Given the following seeds and an input of `[A, B, C, D, E, F]`, performing a single shuffle will result in the following:

|   seed   |       result       |
|----------|--------------------|
|       42 | [B, A, E, F, D, C] |
|      -42 | [E, C, F, A, B, D] |
|        1 | [E, C, A, B, F, D] |
| 11111111 | [B, E, A, D, F, C] |

@MartijnR hope that's enough! Let me know if any more info would be helpful.

MartijnR:
Thank you!

PR #171, opened 2018-03-29
Fix/odk namespace

MartijnR:
Added new ODK namespace preference.

Also slipped in a Changelog, but will move elsewhere if PR gets stuck.


lognaturel (approved):

lognaturel:
I personally don't care so much about mixing things in a PR as long as the commits are clean (as you always make them)! Others may have other thoughts but... they can complain whenever they want! 🎉

PR #179, opened 2018-04-22
Fix/randomize algorithm

### LINKED ISSUE #176: Cross-platform PRNG for randomize()
Continuing on from @ln's comment [here](https://github.com/opendatakit/xforms-spec/pull/167#issuecomment-381241737) and our spec which says about the PRNG to be used:

> linear congruential pseudorandom number generator as the one used by the next() function in Java v7

After further investigation, it seems to me that is not possible to implement Java's PRNG in JavaScript and have a guaranteed equal result. I think this may be related to datatypes (the total space for `long` in java is not available as `number` in javascript) but am not sure. This would be a problem for the scenario where a record is created in e.g. ODK Collect and then edited in Enketo or vice versa, or where data is collected with multiple clients for the same scientific survey.

I'd like to propose to clearly define a simpler PRNG that can be implemented in any language. E.g. something like this: https://gist.github.com/blixt/f17b47c62508be59987b (Park Miller PRNG), but any cross-platform algorithm would work.

I've implemented this here: https://github.com/enketo/enketo-xpathjs/blob/feature/randomize/src/shuffle.js

It changes the test results as follows for array [A,B,C,D,E,F]:

| seed | result |
|------|-------|
| 42 | BDFECA |
| -42 | EDACBF |
| 1 | ECBDFA |
| 11111111 | ADBCEF |

If possible to hold off on publishing a new Collect version with `randomize()`, that would be great.

MartijnR:
closes #176


_sections/30-bindings.md
```diff
@@ -162,7 +162,7 @@ A subset of [XPath 1.0 functions](http://www.w3.org/TR/xpath/#corelib), some fun
...
-<a id="fn:randomize" href="#fn:randomize">`randomize(nodeset arg, number seed)`</a> | Shuffles the nodeset argument using [Durstenfeld's version of the Fisher-Yates algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle). The optional seed argument performs a (reproducable) seeded shuffle using Fisher Yates with a linear congruential pseudorandom number generator as the one used by the [next() function in Java v7](https://docs.oracle.com/javase/7/docs/api/java/util/Random.html#next(int)).
+<a id="fn:randomize" href="#fn:randomize">`randomize(nodeset arg, number seed)`</a> | Shuffles the nodeset argument using the ["inside-out" variant of the Fisher-Yates algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_%22inside-out%22_algorithm). The optional seed argument performs a (reproducable) shuffle using the same algorithm with a _seeded_ Park Miller Pseudo Number Generator.
```
lognaturel: reproducable -> reproducible! 😊

MartijnR: think you!

lognaturel (approved):

PR #180, opened 2018-04-22
Added the `<rank>` form control, closes #104

### LINKED ISSUE #104: Proposal: support for ranking items
Users would like to be able to order a list of items to represent rankings.

One idea for doing this would be to add an `orx:rank` form control element analogous to `select` that also has `item` or `itemset` children. The instance would similarly represent responses to the `rank` question as a <strike>comma-separated</strike> space-separated list of values but these would be ordered and of the same length as the initial list of values rather than an unordered subset.

Of course, I'm open to other implementation ideas!

MartijnR:


lognaturel:
@MartijnR This looks great. If you don't mind, I'd like to wait until there is a concrete implementation before merging. I've gotten the feedback that it's confusing to have things in the spec that don't exist anywhere and since this is coming really soon, I think it's easiest to merge once the feature is in somewhere.

MartijnR:
* [x] ~~add attribute for subset ranking (tbc -> https://forum.opendatakit.org/t/collect-question-type-rank-items/12568/12)~~

MartijnR:
How's it going with the Collect widget? Enketo will have one (a rough one because of minimal funding) next week.

Interesting challenge is the 'empty' state of the widget. Still trying to figure that out UI-wise.

lognaturel:
Great! It's ready on the Collect side as well but probably won't go out for another couple weeks. Clearing the value currently re-establishes the original order -- I couldn't come up with another reasonable option. When you say 'empty' state, do you mean you'd like an enumerator to be able to indicate that they have skipped the question? For that I'd tend to say the form creator should account for that by doing something like asking whether the enumerator has a preferred order and adjusting the relevance of the rank question based on that response.

MartijnR:
> When you say 'empty' state, do you mean you'd like an enumerator to be able to indicate that they have skipped the question?

Yes, and in addition, the enumerator may also decide later to clear the value (the recorded ranked list). This would also make the "required" condition work normally.

So my approach is to visually show an empty state of the widget (which means the value is ""). It may say "Click to start" or something. This way it will have all the functionality that other widgets have (in Enketo). 

In Enketo there will be no way to restore the original order, I think. Clicking reset will simply keep the order but clear the value and visually show the empty state.

lognaturel:
Hmm, we should probably coordinate on this. Do you have a rough sketch of the UI you have in mind?

MartijnR:
Yes, like this:

### No value in model (after reset, or after loading empty form):
screen shot 2018-06-14 at 11 46 52 am


### Has value (and value updates automatically in model, when user changes order):
screen shot 2018-06-14 at 11 46 58 am

MartijnR:
The same 'empty state' requirements could also be met with a 'Confirm' button, but I liked it less.

lognaturel (approved):

lognaturel:
We have implementations ready to ship and XLSForm support!

Issue #151, opened 2017-10-27
Can form submission method be changed to post rather than form-data-post


lognaturel:
From @MartijnR on #149:
> I think method should be in there (but it should actually have value "post" in hindsight). Your question leads me to suspect that ODK Collect is not actually reading/using this attribute. In that case, we could change it to "post" perhaps....

Aggregate expects `form-data-post` but it does seem very odd. I don't know whether there's a reason for it and would be worth more digging. `post` would certainly make more sense. 

- [x] Figure out whether `post` works with current Aggregate (does not)
- [ ] See if there are any breadcrumbs suggesting why `form-data-post` was used

MartijnR:
@lognaturel could Aggregate support both `post` and `form-data-post`, and we document `post` only?

lognaturel:
`post` doesn't currently work with Aggregate but I don't see any issue with also supporting `post`. Still no idea why `form-data-post` was used.

@yanokwa, @ggalmazor, do you see anything I may be missing?

yanokwa:
I have no idea. @wbrunette? Maybe a GAE thing?

lognaturel:
As far as I can tell, `form-data-post` only exists in the ODK world. The type specified in the MIME envelope of submissions is `multipart/form-data` so maybe the idea was to have different `method` "`post` variants" for different kinds of submission formats? I can't think of what those would be and using `post` still seems like the right approach.

lognaturel:
This would also require some coordination with other members of the ecosystem. For example, Ona probably expects `form-data-post`. @ukanga?

wbrunette:
From memory (haven't dug into the code), I thought we were using POST with "multipart/form-data".  This was chosen so you could upload binaries and such without worrying about bad characters and such. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST
https://stackoverflow.com/questions/4526273/what-does-enctype-multipart-form-data-mean
https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4

Since ODK is designed to transmit multiple files in one POST, the multipart/form-data seemed like the best choice. As this enables the submission and pictures/audio/video/etc files to be uploaded in one POST.  

Again going off memory, if you give me pointers to lines of code in repos I am happy to look into the details. I am wondering if "form-data-post" is the same as POST with "multipart/form-data" in whatever library is being used.

lognaturel:
[...]

Issue #23, opened 2015-02-20
add submission element


MartijnR:
@mitchellsundt  wrote:

> If supplied, the one in the form is used. 
> Otherwise, the one in General Settings is used.

(Not supported in Enketo)

MartijnR:
In the mean time, we've added a section about encryption that includes `method`, `action` and `base64RsaPublicKey`. It might be good to make the encryption chapter a section under a new Submission chapter (keeping the original link functional)

An inventory of other submission element's features or outstanding spec issues:

* submission type https://github.com/opendatakit/xforms-spec/issues/151
* auto-send and auto-delete https://github.com/opendatakit/xforms-spec/issues/100

lognaturel:
> It might be good to make the encryption chapter a section under a new Submission chapter (keeping the original link functional)
 
+1

MartijnR:
Continuing on from #100, is the `action` attribute in ODK Collect only used for encrypted data or is it separate from encryption? E.g does`<submission action="https://example.com/submit" method="post">` send a regular unencrypted XML submission to the action URL? 

If ODK Collect doesn't currently do this, I don't feel we need to add this to the spec without a demand for it, but if it does, we can make it describe it like that in the spec. I'm guessing it does.

MartijnR:
And a second question that is coming up while rewriting the submission chapter.

Why are `method/action` required to enable encryption (in particular `action`)? Is this because Aggregate requires using a different submission URL for encrypted records?

Related to the above, any interest in making method/action optional (either both or none), and only `base64RsaPublicKey` required to enable encryption?

lognaturel:
`action` can be used on its own for the form to specify which server it should submit to.

I think the only reason `method/action` are required with encryption is to "save people from themselves". Since encrypted forms are sent as blobs, it's particularly bad to send it to the wrong server.

That said, it's only really an issue if a form with the same `form_id` is present on two servers and basically I think users can be trusted to do the right thing. I'd be for making both method and action optional for encryption unless someone can think of another reason they may be required in the first place. 

@yanokwa, perhaps you remember something else?

yanokwa:
I do not know why `method/action` was required. It always seemed odd, but my guess is that it's by design. @wbrunette do you remember why?

[...]

PR #185, opened 2018-05-30
Added submission chapter, reorganized and edited encryption section 

### LINKED ISSUE #100: Configure auto-send and auto-delete at the form level
**Problem**
We have a user whose campaign has forms that need to be deleted on send because they are sensitive. The same user has other forms that need to be kept on the device because the data is useful to the day to day work of the enumerator.

**Proposal**
In the settings sheet of the XLSForm, we will add an `auto_send` and `auto_delete` heading. If the values underneath this heading is true(), then the form will behave accordingly and override the settings of the device.

I'm not sure how we represent this in an XForm. My first pass is that we use the [submission element](https://www.w3.org/TR/xforms/#submit) and add a namespaced attribute. @MartijnR @mitchellsundt @chrisroberts @meletis Any feedback on this approach?

MartijnR:
closes #100, #23, #151

I'm realizing we're missing some information to implement encryption (the actual submission format, included the encrypted key). Will create a separate issue for that.


lognaturel (approved):

lognaturel:
Nice and clear! Thanks, @MartijnR.

PR #194, opened 2018-06-13
Corrected concat() and once() argument types, closes #191

### LINKED ISSUE #191: concat() args are strings
Args are strings.

Check other functions as well for similar mistake.

MartijnR:


lognaturel (approved):

_sections/30-bindings.md
```diff
@@ -101,7 +101,7 @@ A subset of [XPath 1.0 functions](https://www.w3.org/TR/1999/REC-xpath-19991116/
 
 | function                                  | description |
 |-------------------------------------------|------|
-<a id="fn:concat" href="#fn:concat">`concat(* arg*)`</a> | Deviates from [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-concat) in that it may contain _1 argument_ and that all arguments can be _nodesets_ or strings. It concatenates all string values and _all node values_ inside the provided nodesets.
+<a id="fn:concat" href="#fn:concat">`concat(string arg*)`</a> | Deviates from [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-concat) in that it may contain _1 argument_ and that all arguments can be _nodesets_ or strings. It concatenates all string values and _all node values_ inside the provided nodesets.
```
lognaturel, line 104: Would this be clearer as `string|noseset arg*`?

_sections/30-bindings.md
```diff
@@ -163,7 +163,7 @@ A subset of [XPath 1.0 functions](https://www.w3.org/TR/1999/REC-xpath-19991116/
...
 <a id="fn:distance" href="#fn:distance">`distance(node-set ns | geoshape gs | geotrace gt)`</a> | Returns the distance in meters of either a nodeset of geopoints or a single geoshape value or a single geotrace value (not a combination of these) on Earth, in the sequence provided by the points in the parameter. It takes into account the circumference of the Earth around the Equator and does not take altitude into account.
-<a id="fn:once" href="#fn:once">`once(* calc)`</a> | The parameter will be evaluated and returned if the context nodes's value is empty, otherwise the current value of the context node will be returned. The function is used e.g. to ensure that a random number is only generated once with `once(random())`.
+<a id="fn:once" href="#fn:once">`once(string calc)`</a> | The parameter will be evaluated and returned if the context nodes's value is empty, otherwise the current value of the context node will be returned. The function is used e.g. to ensure that a random number is only generated once with `once(random())`.
```
lognaturel, line 166: `string` to me implies a string literal. Really this is like a [`PathExpr`](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-PathExpr) right? A literal or an XPath expression or a XPath path?

lognaturel (commented):
Had a few thoughts after accepting...

_sections/30-bindings.md
MartijnR: Yes, I think so! I can add that to the next xpath-functions PR.

_sections/30-bindings.md
MartijnR: I think this is correct. Wherever we've added string as the argument type, it can be each of those. It just indicates that the argument will be cast to a string (so the anything-goes type).

E.g. look at `if()` https://www.w3.org/TR/2003/REC-xforms-20031014/slice7.html#fn-if

lognaturel: Ok, makes sense, thanks!

_sections/30-bindings.md
lognaturel: This isn’t merged yet so you can just add it here and then merge!

lognaturel: Though I guess this could also fall under the “anything goes” usage. Either way, you can make the final call and merge.

lognaturel: Though I guess this could also fall under the “anything goes” usage. Either way, you can make the final call and merge.

MartijnR: Ah thanks. I'll merge to save having to fix conflicts with the next one (coming very soon).

PR #195, opened 2018-06-18
Describe compact representation

### LINKED ISSUE #188: Specifying fields to send via SMS
A Long Time Ago™️, a specification was developed by the OpenRosa working group to specify fields that should be sent via SMS and structuring an SMS submission: https://bitbucket.org/javarosa/javarosa/wiki/SMSSendingAPI

ODK Collect users have long wanted this functionality (https://forum.opendatakit.org/t/send-submissions-via-sms/12166/55) and @joeldean is currently making slight modifications to the underlying spec and completing the implementation.

I realized when discussing XLSForm support at https://github.com/XLSForm/pyxform/issues/207#issuecomment-394521060 that SMS support is not included in the spec here. I think it probably should be?

lognaturel:
Closes #188 

This will require some wordsmithing and some references from other sections of the document (e.g. introducing `prefix` and `delimiter` along with `id` and `orx:version`) but I wanted to get the basics as they stand out here so we can make sure the spec itself is ok.

@ukanga @MartijnR if there is already something like this in use in the ecosystem, it would be very helpful to understand whether it's compatible with what we have here.

@jd-alexander @yanokwa please make sure I have consolidated the various ideas and implementations correctly and that nothing is missing.

One big question mark for me is namespaces. The original intent of https://bitbucket.org/javarosa/javarosa/wiki/SMSSendingAPI seemed to be to introduce `tag`, `prefix` and `delimiter` in a new `http://openrosa.org/smsdata` namespace but in practice this was not enforced. My feeling is that this is not worth introducing a new namespace for and that `https://opendatakit.org/xforms` would be appropriate.

### DIFF
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -31,6 +31,9 @@ Nodes inside a primary instance can contain attributes. The client application n
 |---------------|------------
 | `id`          | on the childnode of the primary instance: This is the unique ID at which the form is identified by the server that publishes the Form and receives data submissions. For more information see [this Form List Specification](https://bitbucket.org/javarosa/javarosa/wiki/FormListAPI). \[required\]
 | `orx:version` | on the childnode of the primary instance in the _http://openrosa.org/xforms/_ namespace: Form version which can contain any string value. Like [meta nodes](#metadata) this information is used as a _processing cue_ for the server receiving the submission.
+| `odk:prefix` | on the childnode of the primary instance in the _http://opendatakit.org/xforms_ namespace: optional string prefix which is included at the beginning of the [compact representation](#compact-record-representation-(for-sms))
+| `odk:delimiter` | on the childnode of the primary instance in the _http://opendatakit.org/xforms_ namespace: optional string delimiter which is used to separate prefix, tags and values in the [compact representation](#compact-record-representation-(for-sms))
+| `odk:tag` | on a question node (grandchild of the primary instance) in the _http://opendatakit.org/xforms_ namespace: optional string tag which is used to identify nodes that should be part of the [compact representation](#compact-record-representation-(for-sms))
 | `jr:template` | on any repeat group node in the _http://openrosa.org/javarosa namespace_: This serves to define a default template for repeats and is useful if any of the leaf nodes inside a repeat contains a default value. It is not transmitted in the record and only affects the behavior of the form engine. For more details, see the [repeats](#repeats) section.
 
 The primary instance also includes a special type of nodes for metadata inside the `<meta>` block. See the [Metadata](#metadata) section
--- a/_sections/93-compact-representation.md
+++ b/_sections/93-compact-representation.md
@@ -1,0 +1,61 @@
+---
+title: Compact Record Representation (for SMS) 
+---
+ODK XForms records are generally represented as XML using the structure of the [primary instance](primary-instance). It is also possible to define how a record can be represented more compactly, usually for SMS submission. 
+
+For this representation:
+- The value of the `prefix` attribute on the primary instance's single child is included at the beginning of every record.
+
+- Questions that have a `tag` attribute are represented as the `tag` value followed by the element's value. Questions without a `tag` attribute are omitted.
+
+- The value of the `delimiter` attribute on the primary instance's single child is used to separate components of the compact representation (prefix, tags, values). Defaults to a single space (` `) if not explicitly specified.
+
+Given the following ODK XForm definition:
+
+{% highlight xml %}
+<instance>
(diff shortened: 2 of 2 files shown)


jd-alexander:
@lognaturel Not sure if it's worth mentioning here, but just wondering which attribute are we agreeing on for the form destination number. I am seeing in the old spec it was `action` with a value of `sms://3423223`.

lognaturel:
I think we can add that part to the submission block documentation and I'd rather take that up separately as this set of changes do stand on their own. I think there will need to be some discussion about the `method`.

The original spec introduced the `smspush` method:
`<xforms:submission method="smspush" action="sms://14032030403">`

MartijnR:
Thanks for writing this up!

Re namespaces, the OpenRosa intent was to also namespace all the SMS block nodes themselves (`<person>`, `<firstname>` etc.) and allow nodes outside this block in the regular XForms namespace that would be discarded for the SMS submission, like this:

```xml
<instance>
     <data>
           <sms xmlns="http://openrosa/smsdata" prefix="hh">
                   <name tag="n"/>
           </sms>
           <non-sms><!-- these nodes are all in the regular XForms namespace -->
                  <location/>
           </non-sms>
     </data>
</instance>
```

So, since you've changed the meaning of the `tag` attribute, you've avoided the need to put sms data inside a separate block (and thereby avoided the need for the namespace on the block itself), right?

If so, it makes sense to me to only namespace the 3 new attributes in the OpenRosa or ODK namespace.

Still curious if this spec has any similarities with what was supported by Formhub in the past and is perhaps currently supported by Ona @ukanga. I'll try to see if I can find something.

Update (found so far): 
https://github.com/SEL-Columbia/formhub/issues/1046 
https://github.com/SEL-Columbia/formhub/blob/master/sms_support/tests/fixtures/sms_tutorial.xls

MartijnR:
When converting the above sms_tutorial (formhub) into XForm on xlsform.opendatakit.org: 
[sms_tutorial.xml.txt](https://github.com/opendatakit/xforms-spec/files/2116809/sms_tutorial.xml.txt)

There doesn't appear to be any XForms markup associated with this. When diving deeper into that old GitHub thread it seems this feature may have only been about being able to receive SMS submissions (from ?) and convert them into regular XML submissions upon arrival, using the XLSForm directly to get the special `sms_field`, `sms_option`, `sms_separator`, `sms_keyword`, `sms_date_format`, `sms_datetime_format`, `sms_response`, `sms_allow_medias` information required to make the conversion. Does that sound right @rgaudin? (Hi! ;-))

lognaturel:
> the OpenRosa intent was to also namespace all the SMS block nodes themselves 

Ah yes, thank you, that hadn't occurred to me.

> since you've changed the meaning of the tag attribute, you've avoided the need to put sms data inside a separate block

Exactly. I perhaps should have explained that decision a little more. Using `tag` as a way to include or exclude nodes from the compact representation seems simpler for tools both generating form XML and tools using the resulting representation. Using a block approach like the original spec means that the order of the questions in the instance may not match the order of those questions in the body. It's not a big deal but it does add complexity to form builders (e.g. XLSForm). On the receiving end, optional tags could result in results that are hard to interpret:
- a mix of values with tags and values without tags means the parser needs to be more sophisticated
- blank values without tags really throw things off

The slight downside with always including tags in the compact representation is that it then has to be a bit longer (at least 2 chars per question -- one for the tag and one for the delimiter). But I think the simplicity gained makes it worthwhile. 

I agree with your interpretation of the Columbia work, @MartijnR. It doesn't say anything about how a form could describe a compact representation for SMS sending.

Do we need more information or opinions before introducing `tag`, `delimiter` and `prefix` in the `https://opendatakit.org/xforms` namespace? I'll add that the ODK Technical Steering Committee has approved this direction but we forgot to bring it up here for discussion. Will be better about this in the future!!

MartijnR:
> Do we need more information or opinions before introducing tag, delimiter and prefix in the https://opendatakit.org/xforms namespace?

I think we're good to go! 👍

rgaudin:
>  it seems this feature may have only been about being able to receive SMS submissions (from ?) and convert them into regular XML submissions upon arrival, using the XLSForm directly to get the special information required to make the conversion. Does that sound right @rgaudin?

Hi @MartijnR, you're right ; that's it.

MartijnR:
Thanks!

lognaturel:
This is ready for a more thorough review. I think it can be squashed before merging but wanted to leave the two new commits in so it's easier to see what has changed.

MartijnR (approved):
Looks good. Thanks!

Issue #57, opened 2016-12-22
clarify property() function


MartijnR:
To be done in a way that we can present this as an XPath function (regardless of how exactly it's implemented in JavaRosa or Enketo (it is not)). This is not easy imo.

It would be easy if/once we replace preload items with an external session instance like CommCare because in that case `property(PROP)` is just a shortcut for obtaining `instance('session')/PROP` or something similar.

MartijnR:
@lognaturel writes:
> Sounds like it might be a candidate for deprecation?

Yes, that would be good. And not document it at all perhaps.

lognaturel:
I think removing it from the spec for now is fine. It's ok to have things in JavaRosa that aren't part of the spec. Ideally in the near future JR will get its own documentation that amends this spec with any relevant implementation details and extra features.

When you suggest replacing "preload items with an external session instance like CommCare ", you mean having an external XML document that holds session information and which is retrieved as an external instance as per #50, is that correct?

MartijnR:
What are the exact property() parameter values that JavaRosa supports?

lognaturel:
Oops, totally missed this. [Here](https://github.com/opendatakit/javarosa/blob/3eb73b5c8c1d56b9bd635b1aca6a8e9ab1051fa9/core/src/org/javarosa/xpath/expr/XPathFuncExpr.java#L386) is the relevant code, I believe. It accepts a single string parameter which represents the name of the desired property managed by the `PropertyManager`. I really don't think it belongs in this spec.

And to confirm, you are talking about #50 as a possible way to hold session information, right?

MartijnR:
Thanks. I see [these](https://github.com/opendatakit/javarosa/blob/3eb73b5c8c1d56b9bd635b1aca6a8e9ab1051fa9/core/src/org/javarosa/core/services/properties/JavaRosaPropertyRules.java#L39-L48):

* "DeviceID"
* "cur_locale"
* "logenabled", returning either "Enabled" or "Disabled"
* "jr_openrosa_api"

> I really don't think it belongs in this spec.

Yes, I think I agree. I did coincidentally see a query from Esri about this function and I've asked them for their use case for it (that is not equally well met by preload items). Maybe there is one.

> And to confirm, you are talking about #50 as a possible way to hold session information, right?

Yes

MartijnR:
To be removed as part of #189 fix.

PR #196, opened 2018-06-19
Fix/issue 189

### LINKED ISSUE #189: Clarify Function types, operand conversion, result types
https://opendatakit.github.io/xforms-spec/#fn:coalesce

I tried an integer, which worked. I assume other types would work as well, though I haven't tested.

MartijnR:
Not easy to review this one. The GitHub "rich diff" (per commit) makes it a bit easier.

* described how data types are converted
* added note about dealing with stored booleans
* added return data types
* made some corrections in function descriptions
* categorized the functions

closes #189


MartijnR:
@adammichaelwood, I couldn't add you to reviewers, but it would be great if you could have a look, maybe in particular at the introduction of the XPath Functions section.

_sections/30-bindings.md
```diff
@@ -113,37 +113,33 @@ The functions are categorized based on their main usage. There are no rules asso
...
+<a id="fn:string-length" href="#fn:string-length">`string-length(string arg)`</a> |  number | Deviates from [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-string-length) in that the argument is _required_.
+<a id="fn:contains" href="#fn:contains">`contains(string haystack, string needle)`</a> |  boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-contains).
+<a id="fn:starts-with" href="#fn:starts-with">`starts-with(string haystack, string needle)`</a> | boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-starts-with).
+<a id="fn:ends-with" href="#fn:ends-with">`ends-with(string haystack, string needle)`</a> | boolean | As in [XPath 3.0](https://www.w3.org/TR/xpath-functions-30/#func-ends-with).
+<a id="fn:uuid" href="#fn:uuid">`uuid(number?)`</a> |  string | Without arguments, it returns a random [RFC 4122 version 4](http://tools.ietf.org/html/rfc4122) compliant UUID. With an argument it returns a random GUID with the provided number of characters.
+<a id="fn:property" href="#fn:property">`property(string prop)`</a> |  string | Tbd
```
lognaturel: Can this be quietly removed? Has it ever had a use?

_sections/30-bindings.md
```diff
@@ -97,74 +97,95 @@ Only the _parent_, _child_ and _self_ axes are supported of the [XPath 1.0 axes]
...
+<a id="fn:true" href="#fn:true">`true()`</a> | boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-true).
+<a id="fn:false" href="#fn:false">`false()`</a> |  boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-false).
+<a id="fn:boolean" href="#fn:boolean">`boolean(* arg)`</a> | boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-boolean).
+<a id="fn:boolean-from-string" href="#fn:boolean-from-string">`boolean-from-string(string arg)`</a> | boolean | Deviates from [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice7.html#fn-boolean-from-string) in that it returns `false` for any argument that is not "true" or "1".
+<a id="fn:not" href="#fn:not">`not(boolean arg)`</a> | boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-not).
+<a id="fn:regex" href="#fn:regex">`regex(string value, string expression)`</a> |  string | Returns result of regex test on provided value. The regular expression is created from the provided expression string (`'[0-9]+'` becomes `/[0-9]+/`).
```
lognaturel: Doesn't this return a boolean?

_sections/30-bindings.md
```diff
@@ -97,74 +97,95 @@ Only the _parent_, _child_ and _self_ axes are supported of the [XPath 1.0 axes]
...
+<a id="fn:starts-with" href="#fn:starts-with">`starts-with(string haystack, string needle)`</a> | boolean | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-starts-with).
+<a id="fn:ends-with" href="#fn:ends-with">`ends-with(string haystack, string needle)`</a> | boolean | As in [XPath 3.0](https://www.w3.org/TR/xpath-functions-30/#func-ends-with).
+<a id="fn:uuid" href="#fn:uuid">`uuid(number?)`</a> |  string | Without arguments, it returns a random [RFC 4122 version 4](http://tools.ietf.org/html/rfc4122) compliant UUID. With an argument it returns a random GUID with the provided number of characters.
+<a id="fn:property" href="#fn:property">`property(string prop)`</a> |  string | Tbd
+<a id="fn:Boolean-Functions" href="#fn:Boolean-Functions">**Boolean Functions**</a>|||
+<a id="fn:if" href="#fn:if">`if(boolean condition, string then, string else)`</a> |  string | As in [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice7.html#fn-if).
+<a id="fn:coalesce" href="#fn:coalesce">`coalesce(string arg1, string arg2)`</a> |  string | Returns first non-empty value of arg1 and arg2 or empty if both are empty and/or non-existent.
```
lognaturel: I guess `coalesce` and `once` produce their results by doing something conditional but with their inputs being strings and their outputs being strings it feels odd to have them here.

_sections/30-bindings.md
```diff
@@ -97,74 +97,95 @@ Only the _parent_, _child_ and _self_ axes are supported of the [XPath 1.0 axes]
...
+
+_Note that when an expression result is stored in the XForms model, it is converted to a string by calling the `string()` function. A boolean `false` result, such as from the expression `1 > 2` is stored in the model as the string `"false"`. When referring to that node in another expression as a boolean argument, the **string value** of that node ("true") is converted to a boolean by calling the `boolean()` function which returns the boolean `true` (because `boolean("false") = true()`). To deal with this, it usually best to not do boolean comparisons with stored values (compare strings instead) or use [`boolean-from-string()`](#fn:boolean-from-string)._
+
+The table below describes the functions, and the data types of their arguments and return values, using the following special argument characters:
+
+* `?` argument is optional
+* `*` argument can be repeated
+* `|` alternative argument is allowed 
+
+The functions are categorized based on their main usage. There are no rules associated with these categories and some functions could belong to multiple categories.
```
lognaturel: This categorization seems odd to me. It seems it may have to do with "general feelings of being related to type x?" Why not organize them based on inputs or outputs?

> There are no rules associated with these categories and some functions could belong to multiple categories.

I have trouble parsing this. Do you mean essentially what you said above, that inputs will be cast and so can be of a more specific type (e.g. a function that takes a string can also take a number because it will just be cast to a string)

_sections/30-bindings.md
MartijnR: If used, it's used by Collect users only because it's not supported in Enketo. I never get questions about this though (except Survey123 users - who are using a workaround).

_sections/30-bindings.md
MartijnR: yes, thank you!

_sections/30-bindings.md
MartijnR: I agree it's not a clear-cut case. I chose Boolean because that's where the XForms 1.0 spec put `if()`

_sections/30-bindings.md
lognaturel: Do you know what its intended use was? I don't believe it's used by Collect users (or supported) but perhaps once I know what it's for that will ring a bell.

_sections/30-bindings.md
MartijnR: Yes, exactly. That's what I concluded from the XPath specs as well. The categorization is therefore, more like "where would a developer start looking for this function?" and nothing else (and only for custom functions - referring to the XPath/XForms specs for the categorization of non-custom functions).

> Do you mean essentially what you said above, that inputs will be cast and so can be of a more specific type (e.g. a function that takes a string can also take a number because it will just be cast to a string)

That's exactly it. So unlike what I thought previously the Function Type (Number Type Function or String Type Function or Node Set Type Function) does not influence the evaluation at all. It seems arbitrary. An XPath evaluator has knowledge of the argument types and return type only and casts using `boolean()`, `string()` etc. So indeed a function that requires a string argument will always accept a number argument I think (even INF, -INF), because there doesn't appear to be anything that `string()` cannot cast.

MartijnR: I'll try to re-write this to make it clearer.

_sections/30-bindings.md
MartijnR: It's used to get username, imei, email etc I think.

MartijnR: https://github.com/opendatakit/xforms-spec/issues/57

lognaturel: Ah, ok. Well, I still think it can be removed as we discussed in #57. :)

_sections/30-bindings.md
lognaturel: Ok, that's the best place I can think of for `once` but how about putting `coalesce` in `string`?

MartijnR: I see it as an `if()` with a hardcoded condition, but ... can definitely be argued as String Function too.

lognaturel (approved):
Thanks!

lognaturel:
Will give a day or two to see if anyone else has feedback and then we can merge.

MartijnR:
Seems we're good. Will fix conflict and merge.

PR #199, opened 2018-07-05
Added digest() function, closes #103

### LINKED ISSUE #103: Hashing/digest support
A client wants to be able to compute e.g. an SHA digest over some concatenated answers in a form to produce a secure unique ID.

Could we add something like this: 

https://www.w3.org/MarkUp/Forms/wiki/XPath_Expressions_Module#The_digest.28.29_Function

It doesn't have to be that fancy though.

I'm not sure how you'd want to namespace and I don't really get what the `xf` namespace is or if any of that is supported currently. I leave that up to y'all to suggest.

Thanks!

MartijnR:


lognaturel:
Ohhhh, this guy! Yes, thank you, that's great to have documented.

lognaturel (approved):

PR #200, opened 2018-07-09
Added change process to readme, closes #77

### LINKED ISSUE #77: Document process for proposing changes and/or diverging from the spec
It should be very clear what the recommended process is for proposing new functionality and how to increase the odds that it will be accepted. For example, explicitly mentioning efforts to comply with W3C XForms, describing the decision process for coming up with a signature in the case of a custom XPath function, describing the corresponding XLSForm change... what else should be in that list?

In case the community process doesn't fit someone's needs, it should be clear that the recommendation is to namespace attributes and nodes using their org's namespace. Doing this offers them a path to compatibility if their proposal is eventually approved. More in #73.

I think some of this can be described in the namespace section of the spec and the details probably should be in the README or a separate contribution guide.

MartijnR:
I expect some comments!

E.g. I vaguely remember seeing a template in the forum, but cannot seem to find it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@ ODK XForms Spec
 
 Published at [https://opendatakit.github.io/xforms-spec/](https://opendatakit.github.io/xforms-spec/)
 
-### Contributing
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
This looks right to me! Thanks for formalizing. I don't remember a template!

MartijnR:
Oh, I must have been dreaming. Thanks.

PR #210, opened 2018-10-04
fixed: links to and from encryption sub-spec, closes #209

### LINKED ISSUE #209: Fix links to and from encryption spec

MartijnR:
I hope you've learned not to trust me!

### DIFF
--- a/_sections/100-encryption.md
+++ b/_sections/100-encryption.md
@@ -5,11 +5,11 @@ page: encryption
 
 ### Introduction
 
-This specification is a sub-specification of the [ODK XForms Specification](/). It describes how to encrypt an XForms submission in an manner that is compatible with ODK tools for data aggregation and decryption.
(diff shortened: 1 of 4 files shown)


lognaturel (approved):

lognaturel:
Maybe we should add a PR template to keep ourselves honest! I am not always good about running these either. When I have to check a box that says "I actually ran this", I tend to do it. 😆

MartijnR:
Yes, maybe. In this case, the links worked locally, but the issue was gh-pages serving on a path. I think there must be a way to set such a path in `jekyll serve` to avoid that.

lognaturel:
Oh, I see. Hmm, I wouldn't have known that so even if I had run it locally (which I should have), that probably wouldn't have helped.  ¯\_(ツ)_/¯ Sometimes there's no substitute for just pushing to production and seeing what happens. Good thing you checked the live page because I didn't do that either. 🤕

MartijnR:
Embarrassingly it's still not correct... Sorry. Will send another fix and figure out how to properly run locally to test them.

lognaturel:
🤦‍♀️ 🤦‍♀️  I did check it works locally...

PR #215, opened 2018-10-26
Fix encryption link, compact representation examples and confusing reference to Excel

### LINKED ISSUE #213: decimal-date(): take out Excel reference
https://github.com/opendatakit/docs/pull/893

### LINKED ISSUE #593: When rows attribute is added, use css to size textarea correspondingly
Size is line-height x rows value.

https://github.com/enketo/enketo-xslt/issues/26

lognaturel:
Closes enketo/enketo#593 

I lumped a couple of small changes together hoping that they'll be easy to review commit by commit. 

The middle commit is the interesting one that fixes some things in the section about compact representation where I got some things backwards in the original pass. @MartijnR, @alxndrsn Please make sure I got it right this time.


_sections/93-compact-representation.md
```diff
@@ -30,31 +30,36 @@ Given the following ODK XForm definition:
...
 
-`hh|fn|Mary Kate|ln|Doe`
+`hh+fn+Mary Kate+ln+Doe`
 
-If the delimiter is included in one of the question values, it will be prepended by a slash. For example, the first name `"Mary|Kate"` would be represented as `"Mary\|Kate"`
+If the delimiter is included in one of the question values, it will be prepended by a slash. For example, the first name `"Mary Kate"` would be represented as `"Mary\ Kate"` if the default space delimiter is used.
 
-As in the regular representation, nodes that are not relevant are not included in the compact representation, even if the `tag` attribute is defined.
\ No newline at end of file
+As in the regular representation, nodes that are not relevant are not included in the compact representation. Unlike in the regular representation, nodes that are relevant but empty are not included in the compact representation, even if they have an `odk:tag`.
```
MartijnR, line 65: Just checking. So the argument is that the loss of having a distinction between those relevant empty nodes and irrelevant nodes is worth the gain in having a smaller submission size, right?

lognaturel: Yes, exactly. Should have called that out explicitly in the description. I agree that it's a worthwhile distinction for the full record. As far as I know, this is only used for SMS messages at the moment and they're almost never meant to stand alone. Small size really matters.

I suppose that decision is still not set in stone and could be changed with minimal impact if you think it's not a good way to go. I can't remember whether we explicitly discussed it. And bizarrely I wrote the original example exactly backwards by not including the node in the "normal" submission and including it in the compact submission. I don't know what that was about.

lognaturel: (It can be changed with minimal impact because either users are parsing the results by looking for the tag names or they're using position in which case they expect every node to be populated and hopefully are making the questions required)

MartijnR: I think that's fine. Thanks. Medic Mobile is also not including them in their implementation as @alxndrsn mentions [here](https://github.com/enketo/enketo-core/issues/570#issuecomment-432732607), so let's merge!

MartijnR (approved):

_sections/30-bindings.md
```diff
@@ -172,7 +172,7 @@ For convenience, the functions are categorized based on their main usage. Some f
...
 <a id="fn:format-date-time" href="#fn:format-date-time">`format-date-time(dateTime value, string format)`</a> | string | Returns the provided dateTime value formatted as defined by the format argument using the same identifiers as [`format-date`](#fn:format-date) plus the following:<br/>`%H` 0-padded hour (24-hr time)<br/>`%h` hour (24-hr time)<br/>`%M` 0-padded minute<br/>`%S` 0-padded second<br/>`%3` 0-padded millisecond ticks.\* <br/>\* If form locale can be determined that locale will be used. If form locale cannot be determined the locale of the client will be used (e.g. the browser or app).
 <a id="fn:date" href="#fn:date">`date(* value)`</a> | string | Converts to a string in the ....date format.
-<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | number | Converts dateTime value to the number of days since January 1, 1970 UTC. This is the format used by Excel.
+<a id="fn:decimal-date-time" href="#fn:decimal-date-time">`decimal-date-time(dateTime value)`</a> | number | Converts dateTime value to the number of days since January 1, 1970 UTC.
```
alxndrsn, line 175: :+1: might be good to emphasise **days**, as many might misread this as being a unix timestamp.

_sections/93-compact-representation.md
alxndrsn: That comment of mine's a little misleading because:
1. I was a little confused at the time between irrelevant fields and empty fields
2. we're not actually doing this in production yet, and our implementation will depend on what ends up in the spec!

alxndrsn (approved):
Much clearer, thanks @lognaturel!

_sections/30-bindings.md
lognaturel: https://github.com/opendatakit/xforms-spec/pull/154 Is the discussion on the topic with several other wordings that we tried. Perhaps we can file a separate issue to discuss further clarification?

lognaturel:
It sounds like some folks may want to propose spec changes related to empty nodes. I suggest we first merge this to match previously-accepted reality (the messed up example is pretty confusing) and then proposed changes can be described in forum posts/issues/pull requests as appropriate.

MartijnR:
yes, sounds good to me

PR #218, opened 2019-01-09
clarified dates, dateTimes, today(), and now(), closes #56

### LINKED ISSUE #56: clarify what today() and now() return exactly
@lognaturel writes:
> Would be good to explicitly call out the format and/or say it's DateUtils.roundDate(new Date()), same with now()...

It should be clear what should be returned by these functions at least for datatypes: date, time, datetime, integer, decimal (and string? - that one scares me).

This should be done in a language-agnostic manner (no references to Java or JavaRosa).

MartijnR:
Hi @ggalmazor,

Does this sufficiently and clearly document the date, dateTime, time, datatypes and `today()` and `now()`? 

P.S. I snuck in the missing +, - operators. Sorry about that.


ggalmazor (approved):
It's looking good, although I don't know how to evaluate the added paragraph that says:

> Note that the standard XPath type conversions are extended by this specification in the [`number()` function](#fn:number). This extended functionality provides the ability to perform arithmetic with, and compare, date and dateTime strings.

MartijnR:
Thank you!

Internally an XPath Evaluator will call the `number()` function to cast operands and function parameters to a number when required. It does this e.g. when `+` is used as in `'2' + 1` is evaluated behind-the-scenes via `number('2') + 1`.

I tried to define the ODK behavior with date comparisons `. < today()` and date arithmetic `today() + 1` in the same manner (even though we may use all kinds of wizardry in JavaRosa and Enketo implementations), to make it consistent with native arithmetic and comparisons.

PR #224, opened 2019-03-28
Document odk-instance-first-load

### LINKED ISSUE #223: Document odk-instance-first-load event
Replace every mention of xforms-ready with it.

See [forum discussion](https://forum.opendatakit.org/t/spec-proposal-add-first-load-event-to-replace-xforms-ready/18662) for background.

lognaturel:
Closes #223

I'll update the changelog once this is merged but I also thought it was worth adding a quick note inline for a mythical person who knew the spec well a few months ago and then checked back in. I don't feel strongly about it if reviewers would prefer it were removed. I also maintained the `#event:xforms-ready` anchor.

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -241,15 +241,15 @@ Supported preload attribute combinations are:
 
 | jr:preload    | jr:preloadParams  | value           		| event
 |---------------|-------------------|-----------------------|-------------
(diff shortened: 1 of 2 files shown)


MartijnR (approved):
Thanks. I think it's nice that you care for the mythical persons. In half a year or so, we can remove it.

PR #228, opened 2019-04-02
Document last-saved secondary instances (for auto-fill)

cooperka:
Related PR: https://github.com/opendatakit/collect/pull/2882

cc @lognaturel

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 _site
 *.sublime-project
 *.sublime-workspace
+
+# RubyMine IDE files.
+.idea/
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -123,3 +123,16 @@ The previous section discussed secondary instances with static read-only data th
 {% endhighlight %}
 
 See the [section on URIs](#uris) for acceptable URI formats that refer to an external secondary instance.
+
+### Secondary Instances - Last Saved
+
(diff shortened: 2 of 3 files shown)


lognaturel:
Sorry about the delay getting back to you on this, @cooperka! This feels more tutorial-like than the rest of the specification. I think it'd be great content for user-facing documentation. In particular, we don't mention Collect or any other specific implementation of the spec in this document.

I would tend to add a new section below https://opendatakit.github.io/xforms-spec/#binary-endpoints called something like "Virtual endpoints" and have a table with `jr://instance/last-saved` as the only row for now (with more probably coming - https://forum.opendatakit.org/t/odk-xforms-spec-proposal-introduce-a-virtual-instance-for-metadata/19122).

I think these sentences you had feel appropriate either in the table or below:
Note that the "last-saved" terminology is significant -- it refers to the form instance that was **saved** most recently (as opposed to last-opened or last-finalized, for example).

The most common use case for this feature is to "auto-fill" specific form fields with the last-saved value via [`odk-instance-first-load`](#event:odk-instance-first-load).

Hopefully this makes sense and doesn't feel too nit-picky!

cooperka:
Thanks for the feedback @lognaturel, done.

lognaturel (approved):

lognaturel:
That looks perfect to me! Will give @MartijnR a day or two in case he has feedback on the contents or organization. Thanks so much. 🙏

MartijnR (approved):

MartijnR:
Looks great. Thanks @cooperka @lognaturel !

PR #230, opened 2019-04-22
Add pulldata

lognaturel:
As agreed on at https://forum.opendatakit.org/t/odk-xforms-spec-proposal-add-pulldata-as-custom-xpath-function/18980

I thought about including a note about it existing mostly to query CSVs but decided that might not be necessary at the spec level. Happy to add more if deemed useful, though.

Also please let me know if you have alternate suggestions for the parameter names.


MartijnR (approved):
Looks perfect to me. Thanks!

PR #231, opened 2019-05-20
Add track changes support to audit log

yanokwa:
One might expect that we'd want to support a boolean `true`, but rather we are using the string `"true"`. This approach we are taking is not XPath-friendly, but it is consistent with what we've done with other attributes (namely [auto-send or auto-delete](https://opendatakit.github.io/xforms-spec/#submission-attributes)). @MartijnR I'd love your thoughts here.

Feature discussion: https://forum.opendatakit.org/t/collect-keep-history-of-changes-to-values-in-the-form/15458
Pull request: https://github.com/opendatakit/collect/pull/3072

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -262,3 +262,4 @@ Supported preload attribute combinations are:
 | `odk:location-priority`     | `no-power`, `low-power`, `balanced`, or `high-accuracy` as defined in [LocationRequest](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest). Required to enable location in log.
 | `odk:location-min-interval` | The desired minimum time, in seconds, location updates will be fetched. Required to enable location in log.
 | `odk:location-max-age`      | The maximum time, in seconds, locations will be considered valid. Must be greater than or equal to `odk:location-min-interval`. Required to enable location in log.
+| `odk:track-changes`         | Can be set to `"true"` or `"false"`. If true, whenever an answer is changed, the old value and new value will be added to the log. Attribute is not required and defaults to false.


MartijnR (changes requested):
If this this attribute is not required, would be good to mention that.

Also, in that case, good to mention the default value.

lognaturel:
Agreed with @MartijnR's requests. The naming and namespace seem consistent to me so no concerns there.

To be fully explicit for posterity, my understanding of how booleans work in XPath is that the absence of the attribute or the attribute set to value `""` or the attribute set to a number <= 0 would mean the attribute has boolean false but any other value given to the attribute would mean boolean true. That is, in addition to `odk:track-changes="true"`, `odk:track-changes="false"` would turn on change tracking, as would `odk:track-changes="23"`.

Here, @yanokwa is proposing to continue a convention established [here](https://opendatakit.github.io/xforms-spec/#submission-attributes) with `orx:auto-send` and `orx:auto-delete` to consider `"true"` as boolean true and anything else as false. This is more intuitive so I'm in support of it. If we didn't already have that convention, I might be more hesitant. Note that this is a static value -- expressions are not evaluated.

Somewhat annoyingly, this deviates from [`boolean-from-string`](https://opendatakit.github.io/xforms-spec/#fn:boolean-from-string) which, in addition to `"true"`, considers `"1"` as boolean true.

(@MartijnR I know you got all this, I just want it written down explicitly. Please let me know if I made a mistake, though!)

lognaturel (approved):

MartijnR:
Thanks. I actually missed the question about booleans. I see this as a string value that happens to look a little like a boolean because of the values we accept.  Here are a few more: https://www.ibm.com/support/knowledgecenter/en/SSS28S_8.1.0/XFDL/xfdl_r_xforms_submission.html

I thought an actual boolean attribute would basically be presence (true) or absence (false) of the attribute regardless of any value (in HTML the value can be either "" or the name of the attribute). In this case because there was precedent in XForms/XPath for such a non-evaluated on/off options (`indent`, `omit-xml-declaration`), we chose the same `"true"`, `"false"` string values (for e.g. auto-send). I agree that it makes most sense to consider any other value as irrelevant and therefore use the default.

MartijnR (approved):

MartijnR:
Thanks. I actually missed the question about booleans. I see this as a string value that happens to look a little like a boolean because of the values we accept.  Here are a few more: https://www.ibm.com/support/knowledgecenter/en/SSS28S_8.1.0/XFDL/xfdl_r_xforms_submission.html

I thought an actual boolean attribute would basically be presence (true) or absence (false) of the attribute regardless of any value (in HTML the value can be either "" or the name of the attribute). In this case because there was precedent in XForms/XPath for such a non-evaluated on/off options (`indent`, `omit-xml-declaration`), we chose the same `"true"`, `"false"` string values (for e.g. auto-send). I agree that it makes most sense to consider any other value as irrelevant and therefore use the default.

lognaturel:
Oh good, I'm so happy there's a precedent! Your explanation is perfect.

I'm not sure where I got the < 0 part in my prior comment but 0 or `NaN` being false seems consistent with the [XPath 1](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-boolean) spec. Regardless, I think expecting "true" and considering everything else the default "false" is much easier to explain.

PR #236, opened 2019-06-24
Match formatting of other specs

yanokwa:
Follow on changes to #234 

* Match titlecase that we use everywhere
* Add a link to Events


lognaturel (approved):
Ah, oops. Thank you! 😊

PR #237, opened 2019-06-24
Rename Encryption page for consistency

yanokwa:
The link is called Encryption Algorithms but the filename is called encryption. This has caused me great pain.

The pages render great locally, bu I have not been able to get redirects to work. I think they might not work locally in WEBrick?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
 source "https://rubygems.org"
 
 gem 'jekyll'
+gem 'jekyll-redirect-from'
 gem 'kramdown'
+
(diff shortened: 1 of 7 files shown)


MartijnR (approved):
Thank you! Sorry to hear about the pain! ;)

All looks fine to me. If redirect doesn't work on gh-pages, we can revert. 

P.S. renaming the name of the link to "Encryption", would actually seem slightly preferable to me. But  "Encryption Algorithms" doesn't cause me any pain either, so will leave to you unless @lognaturel jumps in.

yanokwa:
@lognaturel I've given you another PR at https://github.com/opendatakit/xforms-spec/pull/238 that does it the other way. I don't feel strongly, but it is a narrower change and what @MartijnR slightly preferred.

lognaturel:
"I'm with @MartijnR"™️  🇳🇱

PR #241, opened 2019-07-18
Add odk-new-repeat event

### LINKED ISSUE #246: Add odk-new-repeat event
Discussion at https://forum.opendatakit.org/t/xforms-spec-proposal-add-event-fired-when-new-repeat-instance-is-created/19616

In particular, it is added as a custom event rather than [W3C XForms' `xforms-insert`](https://www.w3.org/TR/xforms/#evt-insert) because unlike `xforms-insert`:
- it is narrowly for repeats (ODK XForms currently has no notion of inserting arbitrary nodes)
- its handlers must be nested in the corresponding `repeat` control (W3C XForms identifies the added nodes and the insert position in [event context information](https://www.w3.org/TR/xforms/#fn-event) so handlers can go anywhere)
- it is dispatched separately for each repeat instance (in W3C XForms, a single `insert` action may insert multiple nodes)
<hr>

lognaturel:
Closes #246 

Based on the conversation at https://forum.opendatakit.org/t/xforms-spec-proposal-add-event-fired-when-new-repeat-instance-is-created/19616

Please let me know if it's feeling like too much detail. I tried to capture everything we discussed and leave as little ambiguity as possible. In particular, the contrasting with W3C XForms feels like useful context to me but it's not really necessary.

The example is not realistic -- why in the world would 2 above the user's age be a useful default? My goal is to show an expression and the evaluation context for `value`. If a reviewer thinks a more realistic example would help, I can think of something.

CC @tiritea

### DIFF
--- a/_sections/65-events-and-actions.md
+++ b/_sections/65-events-and-actions.md
@@ -12,8 +12,39 @@ See the W3C XForms specification [section on events](https://www.w3.org/TR/xform
 | --------------------------| ----------- |
 | <a id="event:odk-instance-first-load" href="#event:odk-instance-first-load">`odk-instance-first-load`</a><a id="event:xforms-ready"></a>            | dispatched the first time an instance is loaded |
 | <a id="event:xforms-value-changed" href="#event:xforms-value-changed">`xforms-value-changed`</a>    | dispatched after an instance data node's value changes. |
+| <a id="event:odk-new-repeat" href="#event:odk-new-repeat">`odk-new-repeat`</a>	| dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control. <a href="#the-odk-new-repeat-event">See more</a>.
 
 *Note: `xforms-ready` was previously documented as the event dispatched the first time an instance is loaded. Since that definition does not match the W3C XForms event with the same name, it was deprecated in favor of `odk-instance-first-load`.*
+
+#### The `odk-new-repeat` event
+The `odk-new-repeat` event is dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, `constraints`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control.
+
+Unlike <a href="https://www.w3.org/TR/xforms/#evt-insert">W3C XForms' `xforms-insert`</a>:
+<ul>
+	<li>it is narrowly for repeats (ODK XForms currently has no notion of inserting arbitrary nodes)</li>
+	<li>its handlers must be nested in the corresponding repeat control (W3C XForms identifies the added nodes and the insert position in event context information so handlers can go anywhere)</li>
+	<li>it is dispatched separately for each repeat instance (in W3C XForms, a single insert action may insert multiple nodes)</li>
+</ul>
+
+The `odk-new-repeat` event is never dispatched for repeat instances that are part of the form definition. However, it is dispatched for repeat instances added by evaluation of the `jr:count` attribute value. See <a href="#creation-removal-of-repeats">creation, removal of repeats</a>.
+
+The following example demonstrates giving a node in a repeat a default, user-modifiable value based on other user input:
+
+{% highlight xml %}
+<h:body>
+    <input ref="/data/my_age">
+        <label>Your age</label>
+    </input>
+    ...
+    <repeat nodeset="/data/person">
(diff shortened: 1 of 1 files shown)


_sections/65-events-and-actions.md
```diff
@@ -12,9 +12,40 @@ See the W3C XForms specification [section on events](https://www.w3.org/TR/xform
...
 | <a id="event:xforms-value-changed" href="#event:xforms-value-changed">`xforms-value-changed`</a>    | dispatched after an instance data node's value changes. |
+| <a id="event:odk-new-repeat" href="#event:odk-new-repeat">`odk-new-repeat`</a>	| dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control. <a href="#the-odk-new-repeat-event">See more</a>.
 
 *Note: `xforms-ready` was previously documented as the event dispatched the first time an instance is loaded. Since that definition does not match the W3C XForms event with the same name, it was deprecated in favor of `odk-instance-first-load`.*
 
+#### The `odk-new-repeat` event
+The `odk-new-repeat` event is dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, `constraints`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control.
+
```
MartijnR, line 21: The paragraph below is indeed very good and useful to record for posterity, but I don't think it should be in the spec. Maybe we can at least paste it into the GitHub issue?

_sections/65-events-and-actions.md
```diff
@@ -12,9 +12,40 @@ See the W3C XForms specification [section on events](https://www.w3.org/TR/xform
 | --------------------------| ----------- |
 | <a id="event:odk-instance-first-load" href="#event:odk-instance-first-load">`odk-instance-first-load`</a><a id="event:xforms-ready"></a>            | dispatched the first time an instance is loaded |
 | <a id="event:xforms-value-changed" href="#event:xforms-value-changed">`xforms-value-changed`</a>    | dispatched after an instance data node's value changes. |
+| <a id="event:odk-new-repeat" href="#event:odk-new-repeat">`odk-new-repeat`</a>	| dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control. <a href="#the-odk-new-repeat-event">See more</a>.
 
 *Note: `xforms-ready` was previously documented as the event dispatched the first time an instance is loaded. Since that definition does not match the W3C XForms event with the same name, it was deprecated in favor of `odk-instance-first-load`.*
 
+#### The `odk-new-repeat` event
```
MartijnR: maybe leave out the quotation marks for odk-new repeat as it looks a little odd in the HTML page?

_sections/65-events-and-actions.md
```diff
@@ -12,9 +12,40 @@ See the W3C XForms specification [section on events](https://www.w3.org/TR/xform
 | --------------------------| ----------- |
 | <a id="event:odk-instance-first-load" href="#event:odk-instance-first-load">`odk-instance-first-load`</a><a id="event:xforms-ready"></a>            | dispatched the first time an instance is loaded |
 | <a id="event:xforms-value-changed" href="#event:xforms-value-changed">`xforms-value-changed`</a>    | dispatched after an instance data node's value changes. |
+| <a id="event:odk-new-repeat" href="#event:odk-new-repeat">`odk-new-repeat`</a>	| dispatched when a new instance of a repeat is added to the primary instance and before recomputation of `calculates`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control. <a href="#the-odk-new-repeat-event">See more</a>.
```
MartijnR: I think this can be kept short without the 
> and before recomputation of `calculates`, etc. Actions triggered by `odk-new-repeat` must be nested in the repeat form control

Since that is mentioned right afterwards

MartijnR (changes requested):

MartijnR:
Thanks! This is very good. Sorry, I somehow missed the alert email for this PR.

_sections/65-events-and-actions.md
lognaturel: 👍

_sections/65-events-and-actions.md
lognaturel: Ah yes, good point. 😊

lognaturel:
Thank you! I addressed your feedback in a separate commit so the diff was clear but please squash. And do let me know if you see anything else worth improving.

MartijnR (approved):

PR #243, opened 2019-07-19
Update docs for normalize-space()

### LINKED ISSUE #242: Add documentation for the new `normalize-space` XPath function in Javarosa
A recent pull request in Javarosa for the normalize-space function requires documentation for users. Its definition is here: https://www.w3.org/TR/1999/REC-xpath-19991116/#function-string-length

I recommend it be placed with other string functions at:
https://github.com/opendatakit/xforms-spec/blob/gh-pages/_sections/30-bindings.md#xpath-functions

jkpr:
Closes #242.

I matched what was described for `string-length`.

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -124,6 +124,7 @@ For convenience, the functions are categorized based on their main usage. Some f
 <a id="fn:substring-after" href="#fn:substring-after">`substring-after(string, string)`</a> |  string | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-substring-after).
 <a id="fn:translate" href="#fn:translate">`translate(string, string, string)`</a> |  string | As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-translate).
 <a id="fn:string-length" href="#fn:string-length">`string-length(string arg)`</a> |  number | Deviates from [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-string-length) in that the argument is _required_.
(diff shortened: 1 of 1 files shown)


MartijnR (commented):
Thanks! Would it be problematic to avoid deviating? Especially for XPath 1.0, there are clients that will not deviate and provide different results because they use off-the-shelf or native XPath 1.0 evaluators (Enketo, @tiritea's iOS app).

It would be useful for XPath wizards that do `//item/name[normalize-space() = 'Jane Doe']` or something like that.

jkpr:
Interesting! I am no wizard 🧙‍♂️ so I did not think of that! I was wondering how / why the zero-arg form would be useful.

@lognaturel is planning to get to the zero-arg form eventually. I would do it myself, but I am not certain how to write a good failing test in Javarosa. See my attempt here (down at the bottom is a discussion and link to a git diff on github): https://github.com/opendatakit/javarosa/pull/467

lognaturel:
Agreed, the spec can match XPath 1.0. In the very short term, JavaRosa/Collect may deviate but that's fine.

So in practice, @jkpr, that just means saying "As in XPath 1.0" with the link to the reference. Sorry I didn't think of that decision point!

MartijnR:
Great. 

(Now, of course, I'm wondering if the same could be done for the old string-length()... ;))

lognaturel:
> I'm wondering if the same could be done for the old string-length()

Yes, I was thinking we should clean that up, too. 😊

jkpr:
Am I understanding correctly, then, that I should change the commit to say: 

> As in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-normalize-space)

instead of updating it when we have the zero-arg form?

lognaturel:
Exactly, @jkpr. The specification is the ideal that ties all the different tools in the ODK ecosystem together. Sometimes one tool or another may be slightly out of sync with that ideal and in that case, documentation for that specific tool will reflect that reality.

jkpr:
I didn't know if force updating one commit on this branch would delete this pull request, so I made another pull request: #247.

PR #244, opened 2019-07-19
Removed word binary and added CSV resource, closes #214

### LINKED ISSUE #214: external resources: remove "binary"
XML is not binary

MartijnR:
I was getting confused about the distinction between Virtual and File, and I added some text about that. Let me know if that's not how you see it though!

### DIFF
--- a/_sections/90-uri.md
+++ b/_sections/90-uri.md
@@ -7,7 +7,7 @@ Throughout the XForm format URIs are used to refer to resources outside of the X
 <span id="binary-endpoints" class="invisible-bookmark"></span>
 ### File Endpoints
 
-These endpoints point to files. The following are currently supported:
+These endpoints point to files published in an XForms manifest whose location the client is aware of. The following are currently supported:
 
 | URI format                                 | description
 |--------------------------------------------|----------------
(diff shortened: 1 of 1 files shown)


lognaturel:
I agree that this part of the spec is confusing.

I love everything about this except, as I'm guessing you anticipated, "published in an XForms manifest whose location the client is aware of."

I don't think it's strictly necessary that the files be in a form list manifest (though certainly they should generally be). Also, this spec doesn't go into any OpenRosa details so without context, I don't think it's clear what "manifest" refers to, right?

This is a slightly more involved change but how about focusing on the custom schemes rather than on the endpoints? That is, the two sections would be "file type schemes" and "virtual schemes". For file type schemes, I'm imagining an explanation like "File type schemes allow implementations of this specification to choose how and where they store files of different types used as form media. It is up to clients to resolve URIs prefixed by one of the valid schemes to a file."

Is it the case that "virtual" schemes will always refer to an XML resource? I think so? Where `jr://file` is essentially a path prefix, `jr://last-saved` refers directly to a (dynamically generated) resource and I presume all "virtual" schemes would operate this way as well. So I don't even know if calling them "endpoints" makes sense.

MartijnR:
Thanks!  Sorry, I had to think about this for a few months. 

I agree to leave out reference to manifest, and have done so in an added commit.

I don't think differentiating schemes is the way to go, since I believer on the `jr` part of the URI is the scheme (and it's the same). 

I'm also not sure if the word "endpoint" is the best, but I think it's correct terminology here.

Yes, for now the virtual endpoints are XML, so I think we can leave that until that changes.

lognaturel (approved):
Whoops, yes, what I said about schemes didn’t make sense. I like what you’ve done, thanks!

PR #245, opened 2019-07-19
Removed irrelevant sentence about jr:count, closes #206

### LINKED ISSUE #206: Revisit language about dynamic jr:count
The description of `jr:count` currently includes the following language:

> Note: It is problematic to implement this in a truly dynamic fashion, i.e. when the value changes, to update the number of repeats

This should either be expanded to explain why it's problematic or perhaps changed to a warning. As far as I can tell, a fully-dynamic implementation is fine but some expressions could lead to repeats being added infinitely. That just means a form designer would need to be careful when writing their expressions. 

Related: https://forum.opendatakit.org/t/xls-form-repeat-select-one-question-until-a-is-selected/15318/16

MartijnR:
Am very proud of this one.

### DIFF
--- a/_sections/60-repeats.md
+++ b/_sections/60-repeats.md
@@ -77,7 +77,7 @@ A. Multiple nodes can be defined in the primary instance of the XForm. E.g. see 
 ...
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


lognaturel:
😄 Nobel-prize-worthy if you ask me. I'm still thinking about #244 so I'll go ahead and eat this cake. 🧁

lognaturel (approved):

PR #252, opened 2019-11-11
Fix/repeat ref deviation

MartijnR:
removed devious deviation and updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,29 @@
 
+[2019-11-11]
+--------------------
+##### Removed
+- `xforms-ready` event (replaced by `odk-instance-first-load`).
+- XForms 1.0 deviation to interpret absolute path references to repeat nodes as if they were relative.
+
+##### Added
+- Compact representation of a record: `odk:prefix`, `odk:delimiter`, `odk:tag` attributes
+- Repeat instance naming convention.
+- Record encryption via the `base64RsaPublicKey` attribute.
+- Audit log CSV format.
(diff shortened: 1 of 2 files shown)


lognaturel (approved):

lognaturel:
Thank you! 🎉

More precisely speaking the deviation was with XPath rules, right? It also applied just to paths in a context where a single node was requested but I think losing that nuance is fine.

MartijnR:
> More precisely speaking the deviation was with XPath rules, right? 

Yes

>  It also applied just to paths in a context where a single node was requested but I think losing that nuance is fine.

I don't understand. Could you provide an example please?

lognaturel:
> Could you provide an example please?

Sorry. Words are hard. In the example that sparked all of this linked to from https://github.com/XLSForm/pyxform/issues/393, there's an absolute path in a repeat that's intended to get a nodeset. There has never been a deviation from true XPath there in JavaRosa. The deviation was just in the case where a single node was requested. As I understand it, Enketo ended up with deviation in both because since you don't control the XPath engine, you couldn't change the behavior just in the single node case. So I'd say that the deviation ended up seeming a lot more dire to you than it actually was. Does that make sense? Did I miss something?

MartijnR:
Thanks! Yes, indeed that's totally right. That was an Enketo-only bug caused by the workaround for the single node deviation with was caused by the old bug in JavaRosa. :)

MartijnR:
Ok, and now I really understand, why you mentioned this. It's about the changelog item!! :) 

Sorry. Yes, we can definitely edit that. I'll do that now to make up for my misunderstanding.

PR #257, opened 2019-12-12
Added support for default file values, closes #256

### LINKED ISSUE #256: Add default file values syntax
as agreed at https://forum.opendatakit.org/t/proposal-specify-a-fixed-image-to-be-annotated-or-signed/14954/19

MartijnR:
Enketo should have an implementation in a few days if all goes well. I think it may make sense to do this for all 3 drawing widgets (and that would actually be slightly less work).

### DIFF
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -23,7 +23,7 @@ The _primary instance_ is the first instance defined by the form and should cont
 </instance>
 {% endhighlight %}
 
-Any value inside a primary instance is considered a default value for that question. If that node has a corresponding input element that value will be displayed to the user when the question is rendered.
+Any value inside a primary instance is considered a default value for that question. If that node has a corresponding input element that value will be displayed to the user when the question is rendered. For nodes of [type "binary"](http://localhost:4000/xforms-spec/#data-types), defaults use [file endpoint URIs](http://localhost:4000/xforms-spec/#file-endpoints).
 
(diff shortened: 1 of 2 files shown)


_sections/20-instance.md
```diff
@@ -23,7 +23,7 @@ The _primary instance_ is the first instance defined by the form and should cont
 </instance>
 {% endhighlight %}
 
-Any value inside a primary instance is considered a default value for that question. If that node has a corresponding input element that value will be displayed to the user when the question is rendered.
+Any value inside a primary instance is considered a default value for that question. If that node has a corresponding input element that value will be displayed to the user when the question is rendered. For nodes of [type "binary"](http://localhost:4000/xforms-spec/#data-types), defaults use [file endpoint URIs](http://localhost:4000/xforms-spec/#file-endpoints).
```
lognaturel: Strip the `localhost`, please! 😊

lognaturel (changes requested):
Quick thing!

Collect will get support in v1.26 probably in Feb/March but let's merge now since the spec is agreed on. As written, ALL binary widgets should support this, right? Including audio/video/file? It doesn't seem particularly useful in those cases but seems that all should work?

_sections/20-instance.md
MartijnR: OMG.  Thanks!

MartijnR:
Yes, I think all binary widgets from a spec perspective.

lognaturel (approved):

PR #264, opened 2020-02-07
Added 1.0.0 version

MartijnR:


lognaturel:
Sorry I didn't react to this earlier, @MartijnR. I think it would be helpful to add a note about versioning to the readme since it is so strange to see something that has been around for 10 years without a version suddenly have version 1.0.0. I remember asking for some kind of versioning plan when we agreed to do this strange thing. Do you have thoughts on how it should work moving forward given that we treat the spec as a living document?

MartijnR:
Thanks! Yes, we did discuss that.

We could just be lazy and only add new versions when we know some implementer has a reason for it (as with the repeat-position-injection for Enketo).

But maybe better to increment:

* MAJOR: whenever we change something that could break existing forms (like removing the repeat deviation)
* MINOR: when we add something like a new XPath function, an extra parameter to an XPath function, a new event, a new action, that is very unlikely to break existing forms (because they don't use this feature) - these things can probably be grouped, and don't have to be 'released' immediately, since it's unlikely implementers are using some kind of detection for these.
* PATCH: probably no need to do patch releases for this spec

Pyxform `xforms-version` could be limited to those versions that affect pyxform itself such as setvalue/events and not a new XPath function e.g.. The latter is debatable because it depends on whether ODK Validate is seen as a separate thing or not.

lognaturel (approved):

lognaturel:
Sorry, @MartijnR, I intended to merge this and suggest we evaluate your proposal against the next few changes to see how they'd get versioned and whether we should make any changes to your wording.

> don't have to be 'released' immediately

Would this entail approving PRs but merging them as a cluster when there are several to be released? Or merging but not updating the changelog?

I'm happy for this to be merged now and for us to update the readme with what you've described. We can fine tune as needed.

I think #275 which I just submitted is an interesting case for this versioning plan. I think technically it should be another major version bump but I'm not sure.

MartijnR:
>Or merging but not updating the changelog?

This is what I had in mind, except we would update the changelog but show as 'Unreleased'. But holding off on merging into `gh-pages` is also fine with me. In that case it would be great to merge them into a separate `develop` or `next-release` branch.

PR #275, opened 2020-09-11
Deprecate unique IDs, update deviceID description

### LINKED ISSUE #262: Deprecate simSerial, subscriberID, and phoneNumber
https://forum.opendatakit.org/t/collect-will-need-to-stop-using-imei-as-deviceid-and-making-simserial-and-subscriberid-available/24157/6

lognaturel:
Closes #262

Unique, stable device IDs used to only apply to native applications. Now neither Apple nor Android allow these values to be accessed programmatically for privacy reasons.

Follow up to [this forum discussion](https://forum.getodk.org/t/collect-will-need-to-stop-using-imei-as-deviceid-and-making-simserial-and-subscriberid-available/24157/8?u=ln).

I'm not really sure if/how we should draw attention to this change. It feels weird to quietly change the spec in this way. This is related to the discussion at https://github.com/getodk/xforms-spec/pull/264#issuecomment-647824143 about how to version the specification. Would this be considered a breaking change? It feels like it is because it breaks workflows that depend on the deprecated values. The weird thing is that we had some clients like Enketo that could never satisfy that part of the spec and others like Collect for which it will become impossible to satisfy the requirement past a certain date (Nov 2020). I'm very interested in others' thoughts.


MartijnR:
Thank you! I think this would be good to 'publish' as a 1.1.0 release (if we currently have 1.0.0). Are you thinking we should remove the XForm output from pyxform too, and show a warning/error?

MartijnR (approved):

lognaturel:
Wow, apparently this has sat here approved for 2 years. 😭

An XLSForm warning was introduced in https://github.com/XLSForm/pyxform/pull/473/files.

We did not make an XForms version spec change in XLSForm then. It doesn't really feel to me like the spec itself has changed or that form consumers can do anything useful with a spec version change (especially this late in the game). So I'm going to merge without a version bump for now. If someone feels strongly about a version-related change they can follow up with another PR.

yanokwa (approved):
LGTM

PR #281, opened 2020-12-21
Add quality binding

seadowg:
Add the `odk:quality` for the `binary` type. This is already in use by [ODK Collect](https://docs.getodk.org/form-question-types/#customizing-audio-quality) and [pyxform](https://github.com/XLSForm/pyxform/pull/483).


README.md
```diff
@@ -27,7 +27,7 @@ Pull requests are welcome for any agreed-upon issues! Propose a [specification a
 This specification is not a fixed document and is regularly extended with new features. Occassionally, poor earlier choices are removed, and behavior is clarified or corrected. To propose a small correction or clarification, just open an issue [here](https://github.com/getodk/xforms-spec/issues). For a more significant change in the XForms specification, the following approach should be used:
 
 1. Post a proposal in the [ODK Forum](https://forum.getodk.org) (e.g. under the `Development` category and tagged with `spec-proposal` and `odk-xforms`). If you have a proposed XForm (and/or [XLSForm](https://xlsform.org)) syntax, please include it.
-2. If the ODK community is interested and a syntax candidate emerges, there will be a vote according to the [ODK Voting Rules](https://github.com/getodk/governance/blob/master/GOVERNANCE.md#voting).
+2. If the ODK community is interested and a syntax candidate emerges, there will be a decision made by the [TAB](https://github.com/getodk/governance/blob/master/TAB-GOVERNANCE.md#decision-making).
```
seadowg, line 30: Thought I'd update this while I was in the neighbourhood.

_sections/30-bindings.md
```diff
@@ -2,7 +2,7 @@
 title: Bindings
 ---
 
-A `<bind>` element wires together a primary instance node and the presentation of the corresponding question to the user. It is used to describe the datatype and various kinds of logic related to the data. A bind can refer to any node in the primary instance including repeated nodes. It may or may not have a corresponding presentation node in the [body](#body).
```
seadowg, line 5: Looks like this (and a few other things) are whitespace corrections. Happy to revert them if they seem problematic, but they look right to me.

_sections/80-media.md
```diff
@@ -30,3 +30,12 @@ The `<itext>` element described in the [languages](#languages) section can also
 By default, itext "image" values are not clickable. However, if you also include a "big-image", the image displayed by "image" will be clickable and will display a pannable, zoomable view of the file specified by "big-image". The user interface must provide a way to go back to the form after opening a "big-image". Specifying "big-image" alone has no effect, you must always include "image".
 
 Files referenced by "image" and "big-image" may be the same; however, for performance reasons, it is recommended to create smaller thumbnail images to be referenced by "image".
+
+#### Quality Levels
+
+Audio quality level can be specified using the `odk:quality` [bind attribute](#bind-attributes). Valid quality values are:
+
+* `normal` Should provide a "good enough" quality for most use cases
```
seadowg, line 38: I wanted to just document the intent of each quality level rather than going into detail here as it might be that different clients would need to use different codec etc.

seadowg: We could also link to the Collect docs here, but I'm not sure people would ever look at this while building a form. Thinking maybe it's best to avoid?

lognaturel: I agree that we should not link to a specific client's docs from the spec (we've made an exception for `audit` but if other clients implement it we would redress it).

Would be good to see what @MartijnR thinks about this. Another thing we could do is document file types because it could be hard for downstream tools if clients are not consistent.

_sections/30-bindings.md
lognaturel: I think these kinds of changes make it harder to follow the history and for a spec that might be more important. My preference would be to revert them.

seadowg: Yeah, very good point that it could be a big deal for a spec.

seadowg: Actually I do think it's worth correcting these whitespace errors as they are a problem for Atom with default settings. What I'm going to do is fix introduce a commit before my changes that just adds the whitespace changes, so it's clear in the history that's the change. Let me know if that still feels like a bad idea!

_sections/80-media.md
MartijnR: I would be fine with adding some detail here (even if that may cause a problem for web clients later when they are ready to implement). How about adding file types and adding the codec with parameters (that Collect uses) as an example? 

"in m4a format (e.g. using AAC, mono, 32kHz, 24kbps)"

seadowg:
Closing this for the moment as it's in limbo and not going to become a priority any time soon.

PR #292, opened 2022-03-11
Allow deep linking to individual data types

### LINKED ISSUE #291: Add deep links to data types in documentation
When discussing a function defined in the spec, it's possible to link directly to that function's definition, e.g. https://getodk.github.io/xforms-spec/#fn:uuid

It would be really helpful to be able to do the same with specific data types.

alxndrsn:
Closes #291


lognaturel (approved):
YES. Thank you.

PR #296, opened 2022-06-08
Events & actions: organize all examples together, include setgeopoint examples

eyelidlessness:
Discussed yesterday with @lognaturel: I found the current structure confusing, with one of the event examples under a different section than the others. I also considered renaming the heading to _Actions and Events_, but I'm hesitant as it would break existing off-site links.


lognaturel (approved):
Agreed this is clearer! I thought you were going to add additional subsection for `setgeopoint` -- this is much better than what I understood.

PR #297, opened 2022-06-08
Improve readability on mobile devices/small viewports

eyelidlessness:
- Sets initial page zoom to 1, which is a sensible default for mobile browsers
- Prevents horizontal scrolling of the page by allowing long inline `code` elements to wrap, and wrapping tables in their own scrollable blocks (not dissimilar to the already scrollable code blocks)


lognaturel (approved):
Thanks! I'm sure you are not the only one who was enraged by this. 😄

PR #301, opened 2023-02-02
Document `base64-decode` and process for adding functions

lognaturel:
Documents `base64-decode` as implemented at https://github.com/getodk/javarosa/blob/203f37b702e0c5aadff45d2f7bc7010875328a5f/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L1260

<details>
<summary>Original proposal</summary>
Opening this PR for discussion around two proposed function additions.

I propose adding these in the default/XForm namespace as other custom functions have been introduced in the past. For better or worse, the ODK ecosystem doesn't use namespacing consistently and is not generally compatible with arbitrary XForms. If we do namespace new functions, we either have to expose that to end users of XLSForm or have XLSForm detect namespaced functions and inject namespacing information.

The last commit in this PR describes this approach to function namespacing.

`centroid` is based on http://expath.org/spec/geo#d2e981 but with ODK types.

`base64Binary-to-string` is based on https://www.saxonica.com/html/documentation10/functions/saxon/base64Binary-to-string.html for its simplicity.

Additional candidates:
- http://expath.org/spec/binary#decode-string has more complexity than we need
- https://synapse.apache.org/userguide/xpath.html#base64_decode is similar but the Saxonica function better matches our naming conventions
</details>

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -193,6 +193,7 @@ For convenience, the functions are categorized based on their main usage. Some f
 <a id="fn:Geographic-Functions" href="#fn:Geographic-Functions">**Geographic Functions**</a>|||
 <a id="fn:area" href="#fn:area">`area(node-set ns|geoshape gs)`</a> | number | Returns the calculated area in m2 of either a node-set of geopoints or a geoshape value (not a combination of both) on Earth. It takes into account the circumference of the Earth around the Equator but does not take altitude into account.
 <a id="fn:distance" href="#fn:distance">`distance(node-set ns|geoshape gs|geotrace gt)`</a> |  number | Returns the distance in meters of either a node-set of geopoints or a single geoshape value or a single geotrace value (not a combination of these) on Earth, in the sequence provided by the points in the parameter. It takes into account the circumference of the Earth around the Equator and does not take altitude into account.
+<a id="fn:base64-decode" href="#fn:base64-decode">`base64-decode(base64Binary input)`| `string` | Interprets the contents of the base64Binary value as a sequence of bytes representing a UTF-8 character string and returns the corresponding string.
 
 ### Metadata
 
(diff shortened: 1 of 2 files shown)


_sections/30-bindings.md
```diff
@@ -193,6 +193,9 @@ For convenience, the functions are categorized based on their main usage. Some f
...
 <a id="fn:distance" href="#fn:distance">`distance(node-set ns|geoshape gs|geotrace gt)`</a> |  number | Returns the distance in meters of either a node-set of geopoints or a single geoshape value or a single geotrace value (not a combination of these) on Earth, in the sequence provided by the points in the parameter. It takes into account the circumference of the Earth around the Equator and does not take altitude into account.
+<a id="fn:centroid" href="#fn:centroid">`centroid(node-set ns|geoshape gs|geotrace gt)`| `geopoint` | Returns the mathematical centroid of either a closed node-set of geopoints or a single geotrace value (not a combination of these) as a geopoint.
+<a id="fn:Binary-Functions" href="#fn:Binary-Functions">**Binary Functions**</a>|||
+<a id="fn:base64Binary-to-string" href="#fn:base64Binary-to-string">`base64Binary-to-string(base64Binary input, string encoding)`| `string` | Interprets the contents of the base64Binary value as a sequence of bytes representing a character string in the specified encoding and returns the corresponding string. Supported encodings are UTF-8, UTF-16, and US-ASCII.
```
tobiasmcnulty: Thank you for including this! A couple comments:

- It might make sense to include [string-to-base64Binary](https://www.saxonica.com/html/documentation10/functions/saxon/string-to-base64Binary.html) along with this, since they are sometimes needed in pairs.

- Personally, I find the Saxonica naming convention somewhat confusing; I almost thought this description was for the "encode" operation until I read their examples. If the Saxonica convention is retained, it could make sense to include some examples, similar to their documentation.

> For example, the call saxon:string-to-base64Binary("Dassel", "UTF8") returns the xs:base64Binary value whose lexical representation is "RGFzc2Vs".

> For example, the call saxon:base64Binary-to-string(xs:base64Binary("RGFzc2Vs"), "UTF8") returns the string Dassel.

yanokwa (approved):

PR #302, opened 2023-07-18
Document the new instanceName tag used to customise the form name

mbaudier:
As discussed here:
https://forum.getodk.org/t/collect-manual-instance-naming-will-be-removed-in-v2023-2/40313/7

### DIFF
--- a/_sections/30-bindings.md
+++ b/_sections/30-bindings.md
@@ -212,6 +212,7 @@ The namespace of the meta block is either the default XForms namespace or "https
             <orx:timeEnd/>
             <orx:userID/>
             <orx:instanceID/>
+            <orx:instanceName/>
             <orx:audit/>>
         </orx:meta>
     </data>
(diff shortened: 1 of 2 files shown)


_sections/30-bindings.md
```diff
@@ -227,6 +228,7 @@ The following meta elements are supported:
 | element      | description                                       | default datatype | value             | namespace
 |--------------|---------------------------------------------------|----------------------------------------------------------------------------
 | `instanceID` | The unique ID of the record [required]            | string           | concatenation of 'uuid:' and uuid() | same as meta block
+| `instanceName` | The name of the form. If no binding is specified, defaults to the form title.             | string           |  | same as meta block
```
lognaturel: Thanks so much for making this addition! How about saying "The name of the record" to match `instanceID` above?

I think `If no binding is specified, defaults to the form title.` should be removed because from a spec perspective, if there's no binding, nothing at all is included. Different consumers of the spec can and do make different decisions about what to display if no `instanceName` is provided. ODK Collect does show the form title but ODK Central shows the `instanceID`, for example.

mbaudier: I have pushed the correction!

lognaturel (approved):

lognaturel:
Appreciate you, @mbaudier! À bientôt, j'espère. 😊

mbaudier:
I have spent so much time in this spec that I am happy I could contribute my tiny bit :) Thanks to you all for the great work.

PR #304, opened 2024-01-08
Add Entity updates

lognaturel:
Based on https://docs.google.com/document/d/14s2umYv9x5gBq7Jist94FxUacS2-dE8o4U-DhxVpRK0/edit

### DIFF
--- a/_sections/120-entities.md
+++ b/_sections/120-entities.md
@@ -15,7 +15,7 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 
 | Version  | Changes |
 |----------|-------------------------------------------------------------------------------------------------------------------|
-| 2023.1.0 | Adds Entity updates from form submissions, still with Entities only created on the server                         |
+| 2023.1.0 | Adds Entity updates from form submissions, still with Entities only created or updated on the server              |
 | 2022.1.0 | Adds Entity creation from form submissions, with Entities only created on the server (no offline Entity creation) |
 
 *See section on [Versioning](#versioning)*
@@ -156,7 +156,7 @@ Entity updates are declared in the `entity` element in the [`meta` block](./#met
 - MUST have a `update` attribute populated with a "1" or "true" if the entity should be updated
   - Consumers of submissions that update entities MUST interpret "1" or "true" as indications to update an entity and any other value as indication not to update an entity
 - MUST have a `baseVersion` attribute that is populated with the version of the entity that the form had access to
-- MAY also have a bind to a create attribute as previously defined. In that case, the form designer is responsible for making sure that the id is correctly populated in each case and that the update and create conditions are non-overlapping.
+- MAY also have a bind to a `create` attribute as previously defined. In that case, the form designer is responsible for making sure that the id is correctly populated in each case and that the `update` and `create` conditions don't result in both being truthy at the same time. If both are truthy, the spec consumer processing submissions should do both and one of them will fail.
 - MAY have a direct child label representing a human-readable label
 
 
@@ -167,4 +167,4 @@ The `entities:saveto` [`bind` attribute](./#bind-attributes) declares that the f
 - Property names with `__` prefixes are reserved
 - Property names follow the same rules as form field names ([valid XML identifiers](https://www.w3.org/TR/xml/#NT-Name))
 
-The set of all Entity Properties defined across all forms that populate a specific Dataset define that Dataset’s schema.
+The set of all Entity Properties defined across all forms that populate a specific Dataset define that Dataset’s schema. New properties can be introduced by forms that create or update Entities.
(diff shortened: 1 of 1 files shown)


_sections/120-entities.md
```diff
@@ -3,11 +3,22 @@ title: Entities
...
 ### Introduction
 
-This specification is a sub-specification of the [ODK XForms Specification](./). It describes a semantic layer that identifies the subject of a form ("entity") and its properties. Consumers that implement this specification can process form submissions to extract entity information based on the directives provided in the form definition.
+This specification is a sub-specification of the [ODK XForms Specification](./). It describes a semantic layer that identifies the subject of a form ("Entity") and its properties. Consumers that implement this specification can process form submissions to extract Entity information based on the directives provided in the form definition.
+
+### Versions
+
+| Version  | Changes |
+|----------|-------------------------------------------------------------------------------------------------------------------|
+| 2023.1.0 | Adds Entity updates from form submissions, still with Entities only created on the server                         |
```
ktuite: Should `created` be something more like `modified`? "Still with Entities only modified on the server (no offline Entity updates)"

_sections/120-entities.md
```diff
@@ -89,6 +143,23 @@ Entities are declared in an `entity` element in the [`meta` block](./#metadata)
...
+
+- MUST be a direct child of `meta` in the primary instance.
+- MUST have attribute `id` populated by a [RFC 4122 version 4 UUID](https://www.rfc-editor.org/rfc/rfc4122) representing an existing entity
+  - Consumers of submissions that update entities MUST fail if the `id` attribute does not contain a UUID or if `id` does not reference an existing entity
+- MUST have attribute `dataset` representing the target Dataset for entities updated by submissions of this form
+- MUST have a `update` attribute populated with a "1" or "true" if the entity should be updated
+  - Consumers of submissions that update entities MUST interpret "1" or "true" as indications to update an entity and any other value as indication not to update an entity
+- MUST have a `baseVersion` attribute that is populated with the version of the entity that the form had access to
+- MAY also have a bind to a create attribute as previously defined. In that case, the form designer is responsible for making sure that the id is correctly populated in each case and that the update and create conditions are non-overlapping.
```
ktuite: add ticks to `create`

not sure what "that the update and create conditions are non-overlapping" means...

_sections/120-entities.md
```diff
@@ -89,6 +143,23 @@ Entities are declared in an `entity` element in the [`meta` block](./#metadata)
...
+- MAY have a direct child label representing a human-readable label
+
+
 ### Identifying entity properties
 
 The `entities:saveto` [`bind` attribute](./#bind-attributes) declares that the form field specified by the `nodeset` attribute on the bind should be saved as an Entity Property. The attribute's value is the Entity Property's name and and has the following restrictions:
```
ktuite, line 165: A question that came up from issa was which properties get updated. 

Once you think about adding `saveto`s to the properties you want to update, it starts to comes together, but I think it would be good to explicitly say in the update case that you use `saveto`s to say which properties you want to update and you can actually introduce new entity properties in an update form.

ktuite (approved):

_sections/120-entities.md
lognaturel: It means that the form designer is responsible for making sure that `create` and `update` aren't both truthy. I guess I can say exactly that. Do you think I should say that if they're both truthy, spec consumers should just do both? Maybe we should make the order that Central does them in part of the spec - create then update?

_sections/120-entities.md
lognaturel: Ah yes, I see! I was thinking there's nothing to update locally if Entity creation is only on the server but I see it's a bit strange. I changed it to "only created or updated on the server"

lognaturel:
Thanks so much for the good feedback! I've merged but please consider taking a look and submitting a follow-on PR if you see other opportunities for improvement.

_sections/120-entities.md
```diff
@@ -57,6 +68,49 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
...
+    <h:head>
+        <h:title>Trees circumference update</h:title>
+        <model odk:xforms-version="1.0.0" entities:entities-version="2023.1.0">
+            <instance>
+                <data id="trees_update" version="20240108145123">
+                    <tree/>
+                    <circumference/>
+                    <meta>
+                        <instanceID/>
+                        <entity dataset="trees" id="" update="1" baseVersion="">
```
seadowg, line 85: It initially surprised me that this declares `update="1"` rather than `update="true"`. Reading everything again, I get that you can use `true` or `1`, but maybe it would be better for the `create` and `update` examples to be consistent? I think otherwise there's an implication that `update` is different in someway - for a couple of seconds I was thinking that update conveyed the "next" entity version for instance.

PR #306, opened 2024-02-22
Added clarification about when calculation is performed

sadiqkhoja:
Related: https://github.com/XLSForm/pyxform/issues/546

**Background:** 

Consider a form with two textboxes, second one is calculated. Enketo and Collect both let user to change the value in the second textbox but Collect recalculates the field at the time of submission and Enketo keeps the user entered value.

I can use `xforms-value-changed` to achieve the same behavior in Collect i.e. to recalculate only when first textbox is edited.

We need specification to mention when client is expected to perform calculation.


alxndrsn:
@sadiqkhoja it looks like review of this has been slow, and now there are conflicts.

* if the PR is still relevant, please resolve the conflicts;
* if not, please close

sadiqkhoja:
Closing this as @lognaturel and I discussed long time ago that implementation details shouldn't be added to the specification and when calculation is performed is one such detail.

PR #307, opened 2024-03-12
Add details about baseVersion on update

lognaturel:
I didn't get the lovely word "monotonically" in there but I think this is clear and straightforward.


_sections/120-entities.md
```diff
@@ -159,6 +159,9 @@ Entity updates are declared in the `entity` element in the [`meta` block](./#met
 - MAY also have a bind to a `create` attribute as previously defined. In that case, the form designer is responsible for making sure that the id is correctly populated in each case and that the `update` and `create` conditions don't result in both being truthy at the same time. If both are truthy, the spec consumer processing submissions should do both and one of them will fail.
 - MAY have a direct child label representing a human-readable label
 
+When a consumer of this specification applies an entity `update`, it:
+- MUST treat a `baseVersion` value other than a positive integer as 0
```
seadowg, line 163: Just want to double check we're on the same page here. This is so that we get a version of `1` if we perform an update without a `baseVersion` right?

lognaturel: Yes, exactly. Is this clear enough? The intent is that it include a blank value.

seadowg (approved):

PR #309, opened 2024-03-19
Clarify that repeat naming group must be first child

lognaturel:
The current wording implies that if you have the following structure:

```
repeat
    question
    group
```

the label of that group would be used as the label for repeat instances. I don't believe that was ever the intent. In both Collect and Enketo, a group label is only used as an repeat instance label if it is the first child of that repeat. That's also how it's documented in the ODK docs: https://docs.getodk.org/form-logic/#naming-repeats-to-help-with-navigation


garethbowen (approved):
@lognaturel Looks good to me!

PR #320, opened 2024-10-24
Add v2024.1.0 for offline Entities

### LINKED ISSUE #319: Update entities spec to describe offline behaviour
Proposed wording and discussion: https://forum.getodk.org/t/form-spec-proposal-support-entity-updates-from-multiple-offline-clients/47890

lognaturel:
Closes https://github.com/getodk/xforms-spec/issues/319

I considered different options for leaving breadcrumbs on what v2022.1.0 and v2023.1.0 forms would look like. At first I had examples for all versions but I felt it was too hard to understand. I ended up with notes describing the necessary changes and I'm very interested in feedback on whether that's enough.

Thanks!

### DIFF
--- a/_sections/120-entities.md
+++ b/_sections/120-entities.md
@@ -3,7 +3,7 @@ title: Entities
 page: entities
 ---
 
-Version 2023.1.0
+Version 2024.1.0
 
 ⚠️ In this specification, collections of Entities are referred to as Datasets. The term "Entity List" is generally recommended instead of "Dataset" in text that is intended for users rather than developers.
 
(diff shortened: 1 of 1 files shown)


lindsay-stevens (approved):

_sections/120-entities.md
```diff
@@ -163,6 +171,15 @@ When a consumer of this specification applies an entity `update`, it:
 - MUST treat a `__version` value other than a positive integer, including a missing value, as 0
 - MUST increment the `__version` of its local entity representation by 1 when an update is successfully applied
 
+*Added in spec version 2024.1.0*
+
+The `entity` element for an offline-capable update additionally:
+- MUST have a `branchId` attribute which is populated with a [RFC 4122 version 4 UUID](https://www.rfc-editor.org/rfc/rfc4122) generated by clients with an offline Entity representation.
+  - Clients that do have an offline Entity representation MUST keep the same `branchId` for a given Entity for an entire offline update sequence. They MUST generate a new `branchId` for a new update sequence after receiving a server update.
+  - When a `branchId` attribute is present in a submission, servers SHOULD attempt to process submissions with that same attribute in the correct order
```
lindsay-stevens, line 179: My only question is, after now having more experience with this feature in practice do you think that this (L179) needs to be a `SHOULD` rather than a `MUST`? Like as a ODK spec is there a plausible scenario where this aspect is not complied with.

lognaturel: Thanks for the review!

I put this in mostly for other server implementations -- it's not a critical part of system function and they could choose that it's not all that important because it's not likely to be common. It feels helpful to call out things like this which could be cut from an implementation while still maintaining the spirit of the spec.

If you feel strongly, I'm happy to change it. It doesn't affect any of the core ODK tools.

PR #327, opened 2025-05-12
Document minimum expected file type support for images

lognaturel:
This is for form attachments.

TODO: 
- [x] do Collect/Enketo support SVG? Should it be part of the list? 
- [ ] BMP? TIFF? Does Enketo defer to browser / Collect defer to Android?


latin-panda:
Collect and Enketo support SVG

latin-panda (approved):
LGTM

PR #331, opened 2025-11-23
chore: fix mistake in the phonenumber preload param

garethbowen:
This matches the implementation in [Collect](https://github.com/getodk/collect/blob/60d58edd64d72d0e2247f163cb504ba238288552/metadata/src/main/java/org/odk/collect/metadata/PropertyManager.kt#L31) and [enketo](https://github.com/enketo/enketo/blob/e6a7e3eb68d5f7584885ef42ff3b628b614ce9aa/packages/enketo-core/src/js/form-model.js#L266)


lognaturel (approved):
Thank you!

PR #335, opened 2026-02-03
Entities v2025.1.0: multiple declarations per form

### LINKED ISSUE #330: document entities from repeats
per https://github.com/XLSForm/pyxform/issues/775

lognaturel:
Closes #330 

Based on the spec [discussed on the forum](https://forum.getodk.org/t/odk-xforms-spec-proposal-create-and-update-entities-from-repeats/55164) and in the [ODK brainstorming doc](https://docs.google.com/document/d/1FOw7v3Jr7dUT7NX7XiqCfQtuxiS9ZKD9LLLuV1C9J3I/edit?tab=t.eddsma11418z).

I went back and forth on whether or not to include [the examples](https://forum.getodk.org/t/odk-xforms-spec-proposal-create-and-update-entities-from-repeats/55164/2#p-128997-examples-3). It felt noisy to me and best captured more for implementation purposes but I could easily be convinced to put them in.

I also think [the note about relationships](https://docs.google.com/document/d/1FOw7v3Jr7dUT7NX7XiqCfQtuxiS9ZKD9LLLuV1C9J3I/edit?tab=t.eddsma11418z#heading=h.k6qhflu66fm) isn't necessary here. It was useful when sharing the proposal to be explicit and will be important in user-facing documentation.

### DIFF
--- a/_sections/120-entities.md
+++ b/_sections/120-entities.md
@@ -204,6 +204,4 @@ Clients of this spec MUST associate each `entities:saveto` binding with the near
 
 If an `entities:saveto` binding is declared within a repeat, the nearest ancestor container that has an Entity declaration MUST be that repeat or a descendant group of that repeat. If the nearest ancestor container with an Entity declaration is outside the repeat, the form is invalid.
 
-If an Entity declaration's associated `save_to`s are in a repeat, each node reference in an Entity-level expression (`label`, `create_if`, etc) MUST resolve to a single unambiguous node for each repeat instance. A node reference is valid only if
-* The referenced node is within the repeat subtree, or
-* The referenced node is outside the repeat subtree and is not contained within any repeat that is not an ancestor of the Entity declaration’s repeat.
+If an Entity declaration's associated `saveto`s are in a repeat, each node reference in an Entity-level expression (`label`, `create`, etc) MUST resolve to a single unambiguous node for each repeat instance.
--- a/_sections/120-entities.md
+++ b/_sections/120-entities.md
@@ -15,6 +15,7 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 
 | Version  | Changes |
 |----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 2025.1.0 | Adds support for Entity declarations in groups or repeats.
 | 2024.1.0 | Adds metadata to support clients with offline Entity representations. Clients should only apply creates and updates offline for forms with this version or higher. |
 | 2023.1.0 | Adds Entity updates from form submissions, still with Entities only created or updated on the server                                                               |
 | 2022.1.0 | Adds Entity creation from form submissions, with Entities only created on the server (no offline Entity creation)                                                  |
@@ -23,7 +24,7 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 
 ### Glossary
 
-**Entity**: A uniquely-identified thing that a form is about.
+**Entity**: A uniquely-identified thing that a form can reference, create, or modify.
 
 **Dataset**: A set of Entities of the same type.
 
@@ -36,6 +37,10 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 **User-defined Property**: Properties with arbitrary names defined by the form designer
 
 **Entity Actions**: Actions that can be taken on Entities (`create`, `update`)
+
+**Container**: The root, a group, or a repeat in a form definition.
+
+**Entity declaration**: a `meta/entity` block in a container that declares how Entities are affected by that container when a submission is finalized.
 
 ### Example of an entity-creating form
 
@@ -139,9 +144,9 @@ Consumers MUST reject forms with a version code that is newer than what they can
 
 ### Declaring that a form creates entities
 
-Entities are declared in the `entity` element in the [`meta` block](./#metadata) of the form definition. For entity creation, the `entity` element:
+Entities are declared in the `entity` element in the [`meta` block](./#metadata) of the primary instance root or, starting with spec version 2025.1.0, of a `group` or `repeat`. For entity creation, the `entity` element:
 
-- MUST be a direct child of `meta` in the primary instance.
+- MUST be a direct child of `meta`
 - MUST have attribute `id` populated by a [RFC 4122 version 4 UUID](https://www.rfc-editor.org/rfc/rfc4122)
   - Consumers of submissions that create entities MUST fail to create entities that don't have a UUID `id`
 - MUST have attribute `dataset` representing the target Dataset for entities created from submissions of this form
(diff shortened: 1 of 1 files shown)


lognaturel:
@grzeziek2010, @ktuite, @sadiqkhoja not requiring your review on this but feel free to chime in if you see anything missing or have thoughts on my notes.

_sections/120-entities.md
```diff
@@ -23,7 +24,7 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 
 ### Glossary
 
-**Entity**: A uniquely-identified thing that a form is about.
+**Entity**: A uniquely-identified thing that a form can reference, create, or modify.
```
seadowg, line 27: Good change!

_sections/120-entities.md
```diff
@@ -188,3 +193,17 @@ The `entities:saveto` [`bind` attribute](./#bind-attributes) declares that the f
...
+
+*Added in spec version 2025.1.0*
+
+Any container in a form definition MAY directly include at most one Entity declaration. Nested containers MAY also each directly include at most one Entity declaration.
+
+Clients of this spec MUST associate each `entities:saveto` binding with the nearest ancestor container that has an Entity declaration. If no ancestor container has an Entity declaration, the form is invalid.
+
+If an `entities:saveto` binding is declared within a repeat, the nearest ancestor container that has an Entity declaration MUST be that repeat or a descendant group of that repeat. If the nearest ancestor container with an Entity declaration is outside the repeat, the form is invalid.
+
+If an Entity declaration's associated `save_to`s are in a repeat, each node reference in an Entity-level expression (`label`, `create_if`, etc) MUST resolve to a single unambiguous node for each repeat instance. A node reference is valid only if
```
seadowg: `saveto` rather than `save_to` here.

seadowg (changes requested):

_sections/120-entities.md
```diff
@@ -188,3 +193,17 @@ The `entities:saveto` [`bind` attribute](./#bind-attributes) declares that the f
 - Property names follow the same rules as form field names ([valid XML identifiers](https://www.w3.org/TR/xml/#NT-Name))
 
 The set of all Entity Properties defined across all forms that populate a specific Dataset define that Dataset’s schema. New properties can be introduced by forms that create or update Entities.
+
+### Multiple Entity declarations in a form
+
+*Added in spec version 2025.1.0*
+
+Any container in a form definition MAY directly include at most one Entity declaration. Nested containers MAY also each directly include at most one Entity declaration.
```
sadiqkhoja: it would be nice to link `container` to its definition

```suggestion
Any [container](.#glossary:~:text=Container) in a form definition MAY directly include at most one Entity declaration. Nested containers MAY also each directly include at most one Entity declaration.
```

lognaturel: Fancy text fragment link!

_sections/120-entities.md
lognaturel: Thank you! This piece I originally only had in the XLForm spec. I think it's still helpful here but I removed the bullets about validity because I think they're too restrictive.

seadowg (approved):

_sections/120-entities.md
```diff
@@ -15,6 +15,7 @@ This specification is a sub-specification of the [ODK XForms Specification](./).
 
 | Version  | Changes |
 |----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 2025.1.0 | Adds support for Entity declarations in groups or repeats.
```
lindsay-stevens, line 18: Should this be `2026.1.0`? Both due to the release timing and because the initial version that used a `repeats` column was called `2025.1.0` in the pyxform v4.1.0 release.

_sections/120-entities.md
```diff
@@ -139,9 +144,9 @@ Consumers MUST reject forms with a version code that is newer than what they can
 
 ### Declaring that a form creates entities
 
-Entities are declared in the `entity` element in the [`meta` block](./#metadata) of the form definition. For entity creation, the `entity` element:
+Entities are declared in the `entity` element in the [`meta` block](./#metadata) of the primary instance root or, starting with spec version 2025.1.0, of a `group` or `repeat`. For entity creation, the `entity` element:
```
lindsay-stevens: Suggest using glossary term e.g. "of the form definition." -> "of the container."

_sections/120-entities.md
```diff
@@ -155,9 +160,9 @@ Entities are declared in the `entity` element in the [`meta` block](./#metadata)
 
 *Added in spec version 2023.1.0*
 
-Entity updates are declared in the `entity` element in the [`meta` block](./#metadata) of the form definition. For entity updates, the `entity` element:
+Entity updates are declared in the `entity` element in the [`meta` block](./#metadata) of the primary instance root or, starting with spec version 2025.1.0, of a `group` or `repeat`. For entity updates, the `entity` element:
```
lindsay-stevens: Suggest using glossary term e.g. "of the form definition." -> "of the container."

_sections/120-entities.md
```diff
@@ -188,3 +193,15 @@ The `entities:saveto` [`bind` attribute](./#bind-attributes) declares that the f
...
+
+*Added in spec version 2025.1.0*
+
+Any [container](.#glossary:~:text=Container) in a form definition MAY directly include at most one Entity declaration. Nested containers MAY also each directly include at most one Entity declaration.
+
+Clients of this spec MUST associate each `entities:saveto` binding with the nearest ancestor container that has an Entity declaration. If no ancestor container has an Entity declaration, the form is invalid.
+
+If an `entities:saveto` binding is declared within a repeat, the nearest ancestor container that has an Entity declaration MUST be that repeat or a descendant group of that repeat. If the nearest ancestor container with an Entity declaration is outside the repeat, the form is invalid.
+
+If an Entity declaration's associated `saveto`s are in a repeat, each node reference in an Entity-level expression (`label`, `create`, etc) MUST resolve to a single unambiguous node for each repeat instance.
```
lindsay-stevens: It might be a bit awkward to mention here but I think in regards to the "expression" part I think it's not necessarily that obvious that `single unambiguous node` allows expressions to refer to nodes outside of (above, below, adjacent) the container scope.

_sections/120-entities.md
lognaturel: Forms created with pyxform v4.1.0 would be compatible with this expanded version of the XForms spec! For better or worse, we don't currently version the XLSForm spec.

I think we should leave it because Collect and Central both treat v2025.1.0 as the "multiple Entities in a form" version. Changing it now would mean having to carefully change some changes across all tools.

_sections/120-entities.md
lognaturel: I added a sentence with examples, what do you think?

lindsay-stevens: Thanks! I guess there is also the possibility of using predicates but someone at that level surely would have figured it out based on this wording

lindsay-stevens (approved):

PR #338, opened 2026-02-08
chore: fix broken link to secondary instances section

garethbowen:


lognaturel:
Good find! When external secondary instance support was added, there should have been a general section about all secondary instances, I think. Here's an alternative: https://github.com/getodk/xforms-spec/pull/339

Issue #344, opened 2026-05-04
Incorrect XML namespace in encryption docs


juliemacdonell:
The ODK XForms Specification [docs](https://getodk.github.io/xforms-spec/encryption.html#submission-manifest) on encryption currently say the encrypted submission manifest uses:

`xmlns="http://opendatakit.org/submissions"`

When testing our encrypted surveys, we noticed that submission XMLs for encrypted submissions uses:

`xmlns="http://www.opendatakit.org/xforms/encrypted"`

This appears to be consistent with the encryption util code: https://github.com/getodk/collect/blob/271231bf5b1235a338e23a4f51806d824d065fb0/collect_app/src/main/java/org/odk/collect/android/utilities/EncryptionUtils.java#L88

juliemacdonell:
I actually think there may be a difference here between submissions from Enketo webforms and the ODK Collect app. I just tried a submission from the Enketo webform and the namespace is: `xmlns="http://opendatakit.org/submissions"`. For a subsequent submission from the ODK Collect app for the same survey, the namespace is `xmlns="http://opendatakit.org/submissions"`. What would be needed to have consistent namespaces? In the meanwhile, would it be worth noting in the docs the discrepancy?

lognaturel:
Yes, you're absolutely right that there's a discrepancy. Collect is intended to be the reference implementation so I think we should correct the specification as you've proposed. You can read a bit more on this [in this forum thread](https://forum.getodk.org/t/encrypted-submission-uses-encrypted-namespace-instead-of-submissions-namespace/54552/4?u=ln).

For the core ODK platform, our goal is to replace Enketo with [ODK Web Forms](https://getodk.org/web-forms-preview/) and to reduce a lot of this kind of discrepancy that way. We don't have submission encryption in Web Forms yet but it will be coming very soon.

Would you be willing to share how/why namespace matters in your context? I think the discrepancy was missed because any core ODK tools that do decryption ignore namespace entirely. In practice it doesn't really provide any information that isn't available otherwise as far as I can tell.

juliemacdonell:
@lognaturel So sorry about the very late reply. I somehow missed the notification.

Thanks for the context from the forum thread and sharing the direction about replacing Enketo.

[...]

PR #345, opened 2026-05-04
Correct XML namespace in encryption submission manifest XML docs

juliemacdonell:
See #344


juliemacdonell:
Please see updated comment: https://github.com/getodk/xforms-spec/issues/344#issuecomment-4371888659

_sections/100-encryption.md
garethbowen, line 35: It would be great to update this reference at the same time.

garethbowen (approved):

PR #346, opened 2026-05-06
Add note about unsupported preload attributes

### LINKED ISSUE #326: Clarify that unknown preload param keys should evaluate to empty string
For example, simserial was previously supported but is now deprecated, it should just be empty. Some clients may not support all keys, e.g. phonenumber may not be relevant to all.

lognaturel:
Closes https://github.com/getodk/xforms-spec/issues/326


garethbowen (approved):
👍

PR #347, opened 2026-05-06
Document CSV and geoJSON secondary instances

### LINKED ISSUE #88: Proposal: external CSV data support
Now that we have a way to [add external XML data](https://opendatakit.github.io/xforms-spec/#secondary-instances---external), I'd like to propose to extend this by adding a way to add external CSV data. This proposal aims to meet these 2 requirements:

1. **The CSV data URI is referred to in an XForm to 'load' it** (to avoid magic).
2. **The data can be queried with the full power of XPath** (to avoid requiring a secondary language to use CSV data).

I think we'd end up inventing a CSV query language replacement for XPath if we don't use XPath.

The first part could be met by adding a `jr://file-csv` connector in the same style as `jr://file`, `jr://image` etc:

```xml
<instance id="households" src="jr://file-csv/households.csv"/>
```

The second part could be accomplished by defining a fixed transformation from CSV to XML. I'd like to propose to use the transformation that is identical to the one pyxform performs for the choices sheet when it creates a secondary XML instance for an `<itemset>` which is as follows.

| name | label | rooms | 
|------|-------|-------|
| 0001 | Johnson | 2   | 
| 0034 | Doe   | 5     |

 The above "csv" (presented as a table) is transformed into the following XML (children of `<instance id="..."/>`):

```xml
<root>
    <item>
          <name>0001</name>
          <label>Johnson</label>
          <rooms>2</rooms>
   </item>
   <item>
          <name>0034</name>
          <label>Doe</label>
          <rooms>5</rooms>
    </item> 
</root>
```

P.S. whether these instances are dealt with internally as _actual_ XML Documents or _virtually_ e.g. a database table/document (as CommCare does) is up to the client and not part of the spec.

### LINKED ISSUE #294: Add geojson external secondary instances
https://forum.getodk.org/t/add-support-for-geojson-external-secondary-instance/36114

https://forum.getodk.org/t/ability-to-load-locations-into-the-geowidget-to-guide-data-collection/27406/12

Would make sense to add docs for CSV at the same time: https://github.com/getodk/xforms-spec/issues/88

To verify and add what should happen with:
- [x] leading and trailing spaces for property names and values? This is not consistent across implementations (filed at https://github.com/getodk/xforms-spec/issues/351)
- [x] column or properties names that are invalid XML identifiers?
- [x] duplicate column/properties names?
- [x] nested geojson properties?

lognaturel:
Closes #88
Closes https://github.com/getodk/xforms-spec/issues/294

### DIFF
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -132,7 +132,7 @@ The document referenced by `src` becomes the content of the secondary instance i
 
 It MUST be possible to query any external secondary instance using XPath, the same way as [internal secondary instances](#secondary-instances---internal).
 
-#### Secondary instances from CSV
+##### Secondary instances from CSV
 
 CSV external secondary instances are referenced using the `jr://file-csv/` URI scheme. 
 
@@ -192,7 +192,7 @@ MUST be equivalent for querying to:
 </root>
 {% endhighlight %}
 
-#### Secondary Instances from GeoJSON
+##### Secondary Instances from GeoJSON
 
 GeoJSON secondary instances are referenced using the `jr://file/` URI scheme.
 
--- a/_sections/20-instance.md
+++ b/_sections/20-instance.md
@@ -43,13 +43,15 @@ The primary instance also includes a special type of nodes for metadata inside t
 
 ### Secondary Instances
 
-Secondary instances are used to access read-only data from a form. This data is searchable in XPath. A key use case is in designing _cascading selections_ where the available options of a multiple-choice question can be filtered based on an earlier answer.
+Secondary instances provide read-only access to data that can be queried from a form using XPath expressions. Key use cases include lookup tables and _cascading selections_ where the available options of a multiple-choice question can be filtered based on an earlier answer.
 
-A secondary instance should get a unique `id` attribute on the `<instance>` node. This allows apps to query the data which is outside the root (the primary instance) and would normally not be reachable. The [instance](#fn:instance) function is used for this purpose. For example, `instance('cities')/root/item[country='nl']` filters the items in the ``cities`` instance from the example below.
+Each secondary instance MUST define a unique `id` attribute on the `<instance>` node. The instance data can then be accessed using the [instance](#fn:instance) XPath function. For example, `instance('cities')/root/item[country='nl']` selects all `<item>` elements in the `cities` secondary instance where `<country>` is `nl`.
 
 #### Secondary Instances - Internal
 
-The example below defines two secondary instances with ids `cities` and `neighborhoods`. Internal secondary instances can define [translated strings](#languages) and references to [media](#media).
+Internal secondary instances are embedded directly in the XForm document. They contain static XML data and MAY include references to [translated strings](#languages) and [media](#media).
+
+The example below defines two secondary instances with ids `cities` and `neighborhoods`.
 
 {% highlight xml %}
 <instance>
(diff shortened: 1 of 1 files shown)


lognaturel:
@garethbowen this is a long-standing omission in the spec document. I'm bringing you in for review because it's a good opportunity to verify against what you know of the WF implementation. If you have ready answers to some of the remaining questions on #294 or thoughts on other cases that should be documented, please add to the issue.

_sections/20-instance.md
```diff
@@ -120,10 +120,98 @@ The example below defines two secondary instances with ids `cities` and `neighbo
 
 ### Secondary Instances - External
```
garethbowen: The "Secondary Instances - Internal" header uses h4 not h3. I think this should be consistent, ie "####" (which will require a change to the subsequent headings).

_sections/20-instance.md
```diff
@@ -120,10 +120,98 @@ The example below defines two secondary instances with ids `cities` and `neighbo
...
 
-The previous section discussed secondary instances with static read-only data that is present in the XForm document itself. Another type of secondary instances presents read-only data from an _external_ source. The external source can be static or dynamic and is specified using the additional `src` attribute with a URI value on an empty `<instance>` node. Querying an external instance is done in exactly the same way as for an [internal secondary instance](#secondary-instances---internal).
+Secondary instances can give access to read-only data from an _external_ source. The external source can be static or dynamic and is specified using the additional `src` attribute with a URI value on an empty `<instance>` node. See the [section on URIs](#uris) to learn more about the supported sources.
```
garethbowen: Let's add a definition of "external" in this context...

```suggestion
Secondary instances that give access to read-only data from a source outside of the form XML are called "external". The external source can be static or dynamic and is specified using the additional `src` attribute with a URI value on an empty `<instance>` node. See the [section on URIs](#uris) to learn more about the supported sources.
```

_sections/20-instance.md
```diff
@@ -120,10 +120,98 @@ The example below defines two secondary instances with ids `cities` and `neighbo
...
+{% endhighlight %}
+
+CSV or GeoJSON external instances are converted into an XML-compatible representation with:
+
+* a root element named `root`
+* an `item` element for each CSV row or each GeoJSON `Feature`
+* for CSV, a child of `item` for each column
+* for GeoJSON, a child of `item` for each member of `properties`
+* for GeoJSON, a `geometry` child of `item` containing the feature geometry represented in the ODK format (`Point` to `geopoint`, `LineString` and `MultiLineString` to `geotrace` and `Polygon` and `MultiPolygon` to `geoshape`).
+
```
garethbowen, line 169: As csv and geojson have more points that are different than they have in common, split this list in two, one for each.

garethbowen: In fact, I'd suggest having two separate sections, including splitting the examples below.

_sections/20-instance.md
```diff
@@ -120,10 +120,98 @@ The example below defines two secondary instances with ids `cities` and `neighbo
...
+XML and GeoJSON files are referenced using the `jr://file/` URI scheme. CSV external secondary instances are referenced using the `jr://file-csv/` URI scheme. 
 
 {% highlight xml %}
 <instance id="countries" src="jr://file/country-data.xml"/>
+<instance id="cities" src="jr://file-csv/cities.csv"/>
+<instance id="boundaries" src="jr://file/boundaries.geojson"/>
+{% endhighlight %}
+
+CSV or GeoJSON external instances are converted into an XML-compatible representation with:
```
garethbowen: Is it worth writing a few requirements for CSV parsing? eg

- Each row MUST have at least one column
- Each row MUST have less than or equal to the number of columns of the header row

We could also be prescriptive about what row and column delimiters are valid.

garethbowen (changes requested):
Nice one. Good cleanup job!

A few suggestions inline. The main suggestion is to split up geojson and csv sections as in my head they're distinct concepts.

_sections/20-instance.md
lognaturel: > We could also be prescriptive about what row and column delimiters are valid.

Good ideas. I added a bunch more little things I thought of. 

I didn't end up adding the row constraints but I did reference the RFC that describes CSVs. Let me know if that feels like enough.

Thanks for the nudge to go deeper here, there are a lot of details!

_sections/20-instance.md
```diff
@@ -118,12 +120,138 @@ The example below defines two secondary instances with ids `cities` and `neighbo
...
 {% highlight xml %}
 <instance id="countries" src="jr://file/country-data.xml"/>
 {% endhighlight %}
 
-See the [section on URIs](#uris) for acceptable URI formats that refer to an external secondary instance.
+The document referenced by `src` becomes the content of the secondary instance identified by `countries`. See the [section on URIs](#uris) to learn more about the supported sources.
+
+It MUST be possible to query any external secondary instance using XPath, the same way as [internal secondary instances](#secondary-instances---internal).
+
+#### Secondary instances from CSV
```
garethbowen:
```suggestion
##### Secondary instances from CSV
```

_sections/20-instance.md
```diff
@@ -118,12 +120,138 @@ The example below defines two secondary instances with ids `cities` and `neighbo
...
+  <item>
+    <name>JP-13</name>
+    <label>Tokyo</label>
+    <population>13960000</population>
+    <geometry>35.6895 139.6917 0 0</geometry>
+  </item>
+</root>
+{% endhighlight %}
+
+#### Secondary Instances from GeoJSON
```
garethbowen:
```suggestion
##### Secondary Instances from GeoJSON
```

garethbowen (approved):
Really helpful! Nice one.

I've added two optional suggestions for nesting headings to bring them "under" then external secondary instances header.

Approved whether you accept those or not.

lognaturel:
Thanks for the review, @garethbowen! 🚀

PR #348, opened 2026-05-06
Add new range attributes

### LINKED ISSUE #336: `range` can have `item` or `itemset` child to represent tick labels
Like selects, the range control element can have multiple item children or a single itemset child. Unlike selects, these are optional. When defined, each item defines a labeled tick on the slider:
* The item's **value** defines the position of the tick.
* The item's **label** defines the text displayed at that position.

For each tick label item, its value represents the tick position at which to display the label. It:
* MUST be present.
* MUST parse as a number that is between start and end inclusive.
* MUST align with the step increments:
  * If `odk:tick-interval` is specified, its numeric value MUST be a multiple of the tick-interval value relative to `start`
  * If odk:tick-interval is not specified, its numeric value MUST be a multiple of the step value relative to `start`
* If the no-ticks appearance is specified, the only valid underlying values for items are the values of start and end.

If multiple items have the same value, the client MUST use the first one in document order.

Clients MUST ignore any tick label item that violates the requirements above. If any or all items are invalid, the model value or any other behavior of the range control MUST NOT be affected.

### LINKED ISSUE #337: Add `odk:tick-interval` and `odk:placeholder` attributes to `range`
The `odk:tick-interval` attribute on the `range` control element specifies the spacing between tick marks along the slider track. When present, its value MUST:
* be non-zero
* have an absolute value that is an exact multiple of `abs(step)`
* be less than or equal to `abs(end - start)`

If any of the conditions above are violated or the no-ticks appearance is specified, clients MUST ignore `odk:tick-interval`

The `odk:placeholder` attribute on the range control element defines the value at which the slider is displayed when the question's value is blank/null. Its value must be between `start` and `end`, inclusive, and MUST align with a `step` increment from `start`. If it is not, clients MUST ignore the value.

lognaturel:
Closes https://github.com/getodk/xforms-spec/issues/336
Closes https://github.com/getodk/xforms-spec/issues/337

### DIFF
--- a/_sections/40-body.md
+++ b/_sections/40-body.md
@@ -46,8 +46,8 @@ Within the form controls the following elements can be used:
 | `<label>`     | Child of a [form control](#body-elements) element, `<item>`, `<itemset>` or `<group>` used to display a label. Only 1 `<label>` per form control is properly supported but can be used in [multiple languages](#languages)). As in [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice8.html#ui-commonelems-label) without support for Linking Attributes.
 | `<hint>`      | Child of a [form control](#body-elements) element used to display a hint. Only 1 `<hint>` element per form control is properly supported but can be used in [multiple languages](#languages)). As in [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice8.html#ui-commonelems-hint) without support for Linking Attributes.
 | `<output>`    | Child of a `<label>` or `<hint>` element used to display an instance value, inline, as part of the label, or hint text. It can also be a child of a `<text>` [translation](#languages). As in [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice8.html#ui-output) but only supporting the `value` attribute.
-| `<item>`      | Child of `<select>` or `<select1>` or `<odk:rank>` that defines an choice option. As in [XForms 1.0](https://www.w3.org/TR/2003/REC-xforms-20031014/slice8.html#ui-common-elements-item).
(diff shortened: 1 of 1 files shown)


_sections/40-body.md
```diff
@@ -88,7 +88,31 @@ The following attributes are supported on body elements. Note that most attribut
...
 |`end`            | For the `<range>` element. The upper bound of the range. This attribute is required and its value has to be valid for the data type used.
 |`step`           | For the `<range>` element. The increment between values that can be selected. This attribute is required and its value has to be valid for the data type used.
+| `odk:tick-interval` | For the `<range>` element. Specifies the spacing between tick marks along the slider track. When present, its value MUST be non-zero, its absolute value MUST be an exact multiple of `abs(step)`, and its absolute value MUST be less than or equal to `abs(end - start)`. If any of these conditions are violated, or if the `no-ticks` appearance is specified, clients MUST ignore `odk:tick-interval`.
+| `odk:placeholder` | For the `<range>` element. Defines the value at which the slider is displayed when the question value is blank or null. Its value MUST be between `start` and `end`, inclusive, and MUST align with a `step` increment from `start`. If either condition is violated, clients MUST ignore `odk:placeholder`.
+| `odk:tick-labelset` | For the `<range>` element, see below.  |
```
seadowg: Does this param actually appear in `XForms`? Doesn't `pyxform` just generate `<item>`/`<itemset>` nodes based on this and the discard it?

seadowg (changes requested):

_sections/40-body.md
lognaturel: Yes, absolutely! Thank you, fixed.

seadowg (approved):

lognaturel:
Thanks! Hopefully having this merged will allow any remnants of the attribute we'd once considered (that shall remain unnamed) to leave my brain forever. :D

PR #353, opened 2026-05-08
Clarify entity form processing when both `create` and `update` are true

seadowg:
This should make it clearer what the intention is in this scenario.


lognaturel (approved):

lognaturel:
Thank you!

PR #355, opened 2026-05-26
xml: standardise whitespace

alxndrsn:
* indent with 4 spaces
* remove trailing whitespace


latin-panda (approved):
Approving: Small and safe, I checked other places in the code, and it's using 4 spaces.

Do you know of any lint that can be plugged into this project to flag these issues?

alxndrsn:
> Do you know of any lint that can be plugged into this project to flag these issues?

I had a quick look, but it didn't seem like there was a simple option.

Issue #356, opened 2026-05-26
Hard to build locally


alxndrsn:
Does this repo require Ruby 3.1?

```
/var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/variable.rb:124:in `taint_check': undefined method `tainted?' for "/xforms-spec":String (NoMethodError)

      return unless obj.tainted?
                       ^^^^^^^^^
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/variable.rb:89:in `render'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/block_body.rb:103:in `render_node_to_output'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/block_body.rb:80:in `render'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/template.rb:208:in `block in render'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/template.rb:242:in `with_profiling'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/template.rb:207:in `render'
    from /var/lib/gems/3.2.0/gems/liquid-4.0.3/lib/liquid/template.rb:220:in `render!'
    from /var/lib/gems/3.2.0/gems/jekyll-4.3.1/lib/jekyll/liquid_renderer/file.rb:39:in `block (3 levels) in render!'
    from /var/lib/gems/3.2.0/gems/jekyll-4.3.1/lib/jekyll/liquid_renderer/file.rb:59:in `measure_counts'
    from /var/lib/gems/3.2.0/gems/jekyll-4.3.1/lib/jekyll/liquid_renderer/file.rb:38:in `block (2 levels) in render!'
    from /var/lib/gems/3.2.0/gems/jekyll-4.3.1/lib/jekyll/liquid_renderer/file.rb:63:in `measure_bytes'
    from /var/lib/gems/3.2.0/gems/jekyll-4.3.1/lib/jekyll/liquid_renderer/file.rb:37:in `block in render!'
...
```

Is the `Gemfile.lock` specifically for MacOS?  https://github.com/getodk/xforms-spec/blob/d394121c3591a18fae50c09dfdb61818ce55656f/Gemfile.lock#L62-L64

The GitHub Pages build seems to run on an implied version of Ruby (and perhaps Jekyll).

PR #359, opened 2026-05-26
Gemfile: specify ruby version

alxndrsn:
It seems like some dependencies require Ruby v3.1.  Setting this in the `Gemfile` should help future devs to get their environment set up more easily.

Ref #356


latin-panda (approved):
Approving: low risk. It ensures the app is reproducible on the same safe version, and the version is supported by Jekyll and Kramdown.

PR #364, opened 2026-05-26
xlsform.org: update all links to HTTPS

alxndrsn:
1. https:// is already referenced elsewhere in this repo
2. http:// 301 (permanent) redirects to https://


latin-panda (approved):
Approving: Small safe change

PR #365, opened 2026-05-30
encryption: clarify increment instruction

alxndrsn:
The current wording seems confusing because:

1. "increment X by Y" is more common than "increment X with Y"
2. "at index with 1" might be read as the second (or first) item in the array


garethbowen (approved):
Definitely better.

